PR #6, opened 2020-07-24
Modify pep8speak and minor change for test

fso42:


fso42:
@pep8speaks suggest diff

PR #7, opened 2020-07-27
Alpha beta

matthiasto:
Python alphabeta module to replace the qgis plugin. I tried to compare the outputs of this module with the Qgis pluggin. Some small differences appear. This is probably due to the way the z coordinate is calculated.


fso42:
TODOs:
- [ ] Break out plot function

- [ ] Smaller function

- [ ] pytests for functions

- [ ] Consistent naming -> module name in main functions and logs

- [ ] Use data from slide input example -> Might need preprocessor

- [ ] Move documentation (i.e. Readme) to docs

- [ ] Docstring for each function

PR #30, opened 2020-07-30
moving and clearing SHP2Array to com2AB folder

matthiasto:
also removed the SHPConv.py file


fso42:
Question: would it make sense to move this function to folder (to be created) in in2Trans ?
I think this is a general enough function...

fso42:
and name it in2Trans/shpConversion.py ? Or something similar...

PR #42, opened 2020-08-06
Fix bug ab

matthiasto:
I cleared the code a bit to use dictionaries. 
Added a fullOut flag.
Corrected the try except functions.


docs/moduleIn3Utils.rst
```diff
@@ -85,3 +85,48 @@ Configuration parameters
 :c_ff: standard deviation sigma
 :c_mustart: mean mu - represents upper part of the channel
 :c_muend: mean mu - represents lower part of the channel
+
+
+Read Write Compare raster from or to ASCII files
```
fso42, line 90:
```suggestion
Read / Write / Compare raster from or to ASCII files
```

avaframe/com2AB/com2AB.py
```diff
@@ -251,51 +280,58 @@ def readSplitPoint(fname, header):
...
     ds[0] = 0.0
     dz[0] = 0.0
     angle = np.rad2deg(np.arctan2(dz, ds))
     CuSplit = s[indSplit]
+    # plt.figure(figsize=(10, 6))
+    # plt.plot(s, angle)
+    # plt.show()
     # TODO SPLIT POINT READING
     # get all values where Angle < 10 but >0
     # get index of first occurance and go one back to get previous value
```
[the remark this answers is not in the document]
matthiasto: I am pretty confident it works well to remove all irregularities.
I left the debug blot because I would like to have a look at it on a real topography and see what it gives. Then the three points check would enable us to remove irregularities due to the real topography.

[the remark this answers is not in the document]
fso42: hen make a flag inside the file itself (at the top) debugPlot = False, uncomment the code and put it in a if debugPlot:  construct ... (-> thats a FeatureFlag I was talking about at one point...)

This is for things which we need for now, but will get of rid in future...

avaframe/com2AB/com2AB.py
```diff
@@ -322,23 +340,34 @@ def calcAB(AvaProfile, eqParameters):
 
     s = AvaProfile['s']
     z = AvaProfile['z']
+    distance = s[1] - s[0]
+    delta_ind = int(np.floor(30/distance))
```
fso42, line 344: int(np.floor(30/distance))  gives 0 if distance > 30 

Use np.ceil instead?

matthiasto: So if distance = 40 I still check the next point?

fso42: yes i would say so

PR #43, opened 2020-08-10
A bconfig parser

matthiasto:
Switching to .ini configuration file. 
writing cfig to log file 
creating logging.conf log configuration file (this file can then be used to configure all logging)


avaframe/com2AB/com2ABCfg.ini
```diff
@@ -0,0 +1,62 @@
...
+DGMSource = avaframe/data/avaSlide/slideTopo.asc
+SplitPointSource = avaframe/data/avaSlide/POINTS/slidePoints
+saveOutPath = avaframe/com2AB/Outputs/
+outputName = SlideProfile3
+
+# ProfileLayer          # YOUR PATH TO AVALANCHE PATH
+# DGMSource             # YOUR PATH TO DEM
+# SplitPointSource      # YOUR PATH TO SPLITPOINT
+# saveOutPath           # YOUR PATH TO OUTPUT FILE
+# outputName            # YOUR OUTPUT FILE NAME
```
fso42, line 20: I'd prefer to have the comments for each variable right above it like:

# YOUR PATH TO AVALANCHE PATH
ProfileLayer = ....

# YOUR PATH TO ....

As far as I'm aware this should be no problem for the the configparser lib. Are there any good reasons to not do it like this?

avaframe/com2AB/com2ABCfg.ini
```diff
@@ -0,0 +1,62 @@
...
+# This file is part of Avaframe.
+# This file will be overridden by local_com2ABCfg.ini if it exists
+# So copy this file to local_com2ABCfg.ini, adjust your variables there
+
+
+# Required settings------------------------------
+[INPATH]
+ProfileLayer = avaframe/data/avaSlide/LINES/slideProfiles
+DGMSource = avaframe/data/avaSlide/slideTopo.asc
+SplitPointSource = avaframe/data/avaSlide/POINTS/slidePoints
```
fso42, line 12: I'd like to be able to just give an (general) avalanche directory (according to our input specifications) and alphaBeta then figuress out these files by itself, i.e.:

- Whatever is in LINES are the profiles AB uses
- whatever is in POINTS are the splitpoints
- the one ASCII that is in the avalanche root directory is the DGM

and so on. 

If we feel this is too uncertain, i.e. maybe we want points for something else, we could define that the user needs to name it with _AB_ somewhere in the filename, so we check the directory for files with this filename...

This means these variable don't nee to be set here, same goes for the saveOutPath, as this is just AvaLancheRoot/Output or WorkDir/Output

I suggest to also don't provide outputName, but let it be something like
AvalanchenameNameofprofile  
where nameofprofile comes from the attribute in the slideProfiles file or, if not provided there, we use a default ProfileX naming...

avaframe/configLogger.py
```diff
@@ -0,0 +1,11 @@
...
+class ConfigLogger(object):
+    def __init__(self, log):
+        self.__log = log
+    def __call__(self, config):
+        self.__log.info("Config:")
+        config.write(self)
+    def write(self, data):
+        # stripping the data makes the output nicer and avoids empty lines
+        line = data.strip()
+        self.__log.info(line)
```
fso42, line 11: Maybe move the whole logging setup also to in3Utils? Keeps our root directory cleaner...

avaframe/logging.conf
```diff
@@ -0,0 +1,32 @@
...
+level=INFO
+args=(sys.stdout,)
+
+[handler_fileHandler]
+class=handlers.TimedRotatingFileHandler
+interval=midnight
+backupCount=5
+formatter=simple
+level=DEBUG
+args=('%(logfilename)s',)
```
fso42, line 32: see note on configLogger.py -> move to in3Utils

fso42, line 32: This is a setup suitable for a constantly running service. 
I suggest we use a simple overwrite on the _logfilename_ file

avaframe/runCom2AB.py
```diff
@@ -2,44 +2,74 @@
...
+cfgPath = cfg['INPATH']
+cfgsetup = cfg['ABSETUP']
+cfgFlags = cfg['FLAGS']
+
+#########################################
+# Load all input Parameters for logging #
+#########################################
+logFileName = os.path.join(cfgPath['saveOutPath'], "runCom2AB.log")
+if os.path.exists(logFileName):
+    os.remove(logFileName)
```
fso42, line 36: See my note on using an overwrite fileHandler for the log file, then these lines are not needed anymore

avaframe/runCom2AB.py
```diff
@@ -2,44 +2,74 @@
...
+logging.config.fileConfig(fname='avaframe/logging.conf', defaults={'logfilename': logFileName}, disable_existing_loggers=False)
 log = logging.getLogger(__name__)
+config_logger = ConfigLogger(log)
+config_logger(cfg)
```
fso42, line 40: This could be moved to configLogger.py, so one would just have to call
log = configLogger.initiateConfig(logFileName) 
and
configLogger.logCfg(cfg)

avaframe/com2AB/com2AB.py
```diff
@@ -245,6 +246,33 @@ def findSplitPoint(AvaProfile, splitPoint):
...
+
+def readABinputs(cfgINPATH):
+    cfgAva = cfgINPATH['AvaLancheRoot']
+    cfgPath = {}
+
+    ProfileLayer = glob.glob(cfgAva + '/Inputs/LINES/*.shp')
+    cfgPath['ProfileLayer'] = ''.join(ProfileLayer)
+
+    DGMSource = glob.glob(cfgAva + '/Inputs/*.asc')
+    cfgPath['DGMSource'] = ''.join(DGMSource)
```
fso42, line 258: Here we might have a potential Problem if more then 1 .asc files are present. 
We could either just take the first one, or throw an error if more than one .asc files are present (I tend more towards the second option...

avaframe/in3Utils/logging.conf
```diff
@@ -24,9 +24,9 @@ level=INFO
 args=(sys.stdout,)
 
 [handler_fileHandler]
-class=handlers.TimedRotatingFileHandler
+class=FileHandler
 interval=midnight
 backupCount=5
```
fso42: these 2 options are not needed anymore, right?

avaframe/runCom2AB.py
```diff
@@ -24,17 +24,15 @@
...
 cfgFlags = cfg['FLAGS']
 
 #########################################
 # Load all input Parameters for logging #
 #########################################
 logFileName = os.path.join(cfgPath['saveOutPath'], "runCom2AB.log")
-if os.path.exists(logFileName):
-    os.remove(logFileName)
-logging.config.fileConfig(fname='avaframe/logging.conf', defaults={'logfilename': logFileName}, disable_existing_loggers=False)
+logging.config.fileConfig(fname='avaframe/in3Utils/logging.conf', defaults={'logfilename': logFileName}, disable_existing_loggers=False)
```
fso42, line 35: What is the purpose of the 
disable_existing... flag? Do we need it?

avaframe/com2AB/com2AB.py
```diff
@@ -245,6 +246,33 @@ def findSplitPoint(AvaProfile, splitPoint):
...
+
+    SplitPointSource = glob.glob(cfgAva + '/Inputs/POINTS/*.shp')
+    cfgPath['SplitPointSource'] = ''.join(SplitPointSource)
+
+    saveOutPath = os.path.join(cfgAva, 'Outputs/')
+    if not os.path.exists(saveOutPath):
+        # log.info('Creating output folder %s', saveOutPath)
```
fso42, line 265: commented on purpose?

avaframe/com2AB/com2AB.py
```diff
@@ -245,6 +246,33 @@ def findSplitPoint(AvaProfile, splitPoint):
...
+        os.makedirs(saveOutPath)
+    cfgPath['saveOutPath'] = saveOutPath
+
+    DefaltName = str(cfgAva).split('/')[-1]
+    cfgPath['DefaltName'] = DefaltName
```
fso42, line 270: Not a functional issue, just correction: I'd change all to defaultName  (with the u)

avaframe/runCom2AB.py
matthiasto: If you want to write the results in another logger, you need this flag. It is set to True as default. Since we only have one logger, it is fine to remove it, it was just in case

avaframe/com2AB/com2AB.py
matthiasto: I create the logger after this line so this is not printed anyway... I have to find a way to make it work.

fso42: I think the log could end up in the root of the avalanche , so this would make it easier, right?

fso42: Mh, i think the log file can end up in the avalanche root, so this would make things easier, right?

fso42:
Pytest runs, however trying to run runcom2AB.py in a clean environemnt results in:
```
Traceback (most recent call last):
  File "runCom2AB.py", line 27, in <module>
    cfgPath = com2AB.readABinputs(cfg['INPATH'])
  File "/home/felix/.conda/envs/avaframe/lib/python3.8/configparser.py", line 960, in __getitem__
    raise KeyError(key)
KeyError: 'INPATH'
```
I'm looking into it...

fso42:
Merging it as is, will open a new branch right away as my restructuring is a bit too much for this PR...

PR #44, opened 2020-08-12
Restructure configuration and log handling

fso42:
Please have a look at `runCom2AB.py`
I overhauled configuration reading and log handling. 
Please give it a try and let me know what you think of it.


avaframe/com2AB/com2AB.py
```diff
@@ -247,32 +249,32 @@ def findSplitPoint(AvaProfile, splitPoint):
...
-    ProfileLayer = glob.glob(cfgAva + '/Inputs/LINES/*.shp')
-    cfgPath['ProfileLayer'] = ''.join(ProfileLayer)
+    profileLayer = glob.glob(cfgAva + '/Inputs/LINES/*.shp')
+    cfgPath['profileLayer'] = ''.join(profileLayer)
 
-    DGMSource = glob.glob(cfgAva + '/Inputs/*.asc')
+    dgmSource = glob.glob(cfgAva + '/Inputs/*.asc')
     try:
-        assert len(DGMSource)==1, 'There should be only one and only one DEM .asc file in ' + cfgAva + '/Inputs/'
+        assert len(dgmSource)==1, 'There should be only one DEM .asc file in ' + cfgAva + '/Inputs/'
```
matthiasto, line 261: There should be one and only one DEM... This message stands if there is no DEM file.
We should use DEM and not DGM i think

avaframe/in3Utils/logUtils.py
```diff
@@ -0,0 +1,48 @@
+""" Defining a writable object to write the config to the log file"""
+import logging
+import logging.config
+import os
+from datetime import datetime
+
+def dumpCfg(cfg, cfgName='Unspecified'):
+    '''Dumps a configParser objecct to log file'''
```
matthiasto, line 8: 'object'

avaframe/com2AB/com2AB.py
matthiasto, line 261: There should be one and only one DEM... This message stands if there is no DEM file.
We should use DEM and not DGM i think

avaframe/in3Utils/logUtils.py
matthiasto: I am personally not found of the word 'Dump'. I find it a bit aggressive and code should be neutral. I think we should use something like Cfg2Log or writeCfg for example.

avaframe/in3Utils/logUtils.py
```diff
@@ -0,0 +1,48 @@
+""" Defining a writable object to write the config to the log file"""
+import logging
+import logging.config
+import os
+from datetime import datetime
+
+def dumpCfg(cfg, cfgName='Unspecified'):
```
matthiasto, line 7: I am personally not found of the word 'Dump'. I find it a bit aggressive and code should be neutral. I think we should use something like Cfg2Log or writeCfg for example.

avaframe/runCom2AB.py
```diff
@@ -1,74 +1,67 @@
...
 
 log.info("Running com2ABMain model on test case DEM \n %s \n with profile \n %s ",
-         cfgPath['DGMSource'], cfgPath['ProfileLayer'])
+         cfgPath['dgmSource'], cfgPath['profileLayer'])
 
-DGM = com2AB.readRaster(cfgPath['DGMSource'])
-Avapath = com2AB.readAvaPath(cfgPath['ProfileLayer'], cfgPath['DefaltName'], DGM['header'])
-SplitPoint = com2AB.readSplitPoint(cfgPath['SplitPointSource'], DGM['header'])
+# Read input data for ALPHABETA
+dgm = com2AB.readRaster(cfgPath['dgmSource'])
```
matthiasto, line 51: dem

avaframe/com2AB/com2AB.py
awirb: I also think we should stick to english and use DEM

PR #49, opened 2020-08-14
Add com1DFA files

awirb:


avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -0,0 +1,30 @@
...
+flagVarMU = False
+
+
+[AIMEC]
+# directory to store results for Aimec
+aimecDir = Outputs/com1DFA/
+# pressure limit for Aimec to perform computations
+calcPressureLimit = 10
+# width of the avalanche path domain for computations in Aimec
+domainWidth = 600
```
fso42, line 30: I guess these will be moved to the aimec module?

matthiasto: Not sure for the aimecDir = Outputs/com1DFA/ because you create the data in com1DFA and need to know where to export the data.
What is sure is that you don't need the calcPressureLimit and domainWidth any more and you don't need to create the aimecPathFile anymore. I think this will be done with aimec

fso42:
> Looks good to me!
> 
> Should we in every module add some time info about how long it took to run the module?
> 
> startTime = time.time()
> bla bla
> endTime = time.time()
> log.info(('Took %s seconds to calculate.' % (endTime - startTime)))

For now? 
I would say no, as it is way more easy, flexible, precise and does not need any maintenance on our side to just fire up a profiler. 

If we need simple timing info, we can also just use timeit or similar...

avaframe/com1DFA/com1DFACfg.ini
awirb: The aimec stuff will be moved to ana3AIMEC as a next step, so the com1DFA output is saved to Outputs/com1DFA and there will be a helper function that exports the required data from there.

awirb: Shall we first move the things from Aimec or merge it as is and do the changes this in a new feature branch?

fso42: Merge it as is and do the changes in a new branch...

PR #65, opened 2020-08-18
Streamline  and simplifiy generateTopo

fso42:
There are still some todos in the code that need to be addressed before merging


awirb:
I cleaned up the TODOs and simplified the getReleaseArea.py function a bit more

PR #69, opened 2020-08-20
Add Tests

awirb:


fso42 (changes requested):
Please clean out the logs and the pngs....

Maybe add .log and pngs to gitignore...

PR #73, opened 2020-08-24
Add topo pytest

awirb:


matthiasto (changes requested):
run the autopep to get rid of the falke 8 issues and then its good

PR #76, opened 2020-08-25
Test aimec

matthiasto:


avaframe/in2Trans/geoTrans.py
```diff
@@ -362,11 +365,12 @@ def path2domain(xyPath, w, header):
     y = xyPath['y']
     w = w/2/csz
     # Shift grid origin to (0,0)
-    x -= xllcenter
-    y -= yllcenter
```
awirb, line 366: I would suggest that we stick to 'full' operations like x = x +  1 instead of x +=  1

awirb (changes requested):

PR #81, opened 2020-08-25
Input data

matthiasto:
Adding Lines and Points in inputs (except for flat and inclined plane)


awirb (changes requested):
Could you make path Aimec and path AB identical where they both exist and the paths for the topos with and without the channels could also be identical I think regarding their location

PR #84, opened 2020-08-27
Add quick plot function

awirb:
* Add quick plot function to compare two datasets
* Add functions to handle files and directories and use this within the modules
* Add plot flag in general configuration file
* Add Outputs folder and xml files to gitignore in avaframe folder

### DIFF
--- a/avaframe/avaframeCfg.ini
+++ b/avaframe/avaframeCfg.ini
@@ -4,9 +4,12 @@
 
 [MAIN]
 # Path to avalanche directory
-avalancheDir = data/avaHockey
+avalancheDir = data/avaHockeySmoothSmall
 
 
 [FLAGS]
 # True if plots shall be plotted to screen
 showPlot = False
+
+# True if comparison to reference shall be performed
+compRef = True
--- a/avaframe/avaframeCfg.ini
+++ b/avaframe/avaframeCfg.ini
@@ -4,4 +4,9 @@
 
 [MAIN]
 # Path to avalanche directory
-avalancheDir = data/avaHockeySmoothChannel
+avalancheDir = data/avaHockey
+
+
+[FLAGS]
+# True if plots shall be plotted to screen
+showPlot = False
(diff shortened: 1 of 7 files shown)


avaframe/avaframeCfg.ini
```diff
@@ -4,4 +4,12 @@
...
-avalancheDir = data/avaHockeySmoothChannel
+avalancheDir = data/avaHockeySmoothSmall
+
+
+[FLAGS]
+# True if plots shall be plotted to screen
+showPlot = False
+
+# True if comparison to reference shall be performed
+compRef = True
```
matthiasto, line 15: should this be here? We are the only ones who will compare to the reference no?

avaframe/in3Utils/fileHandlerUtils.py
```diff
@@ -0,0 +1,143 @@
...
+# change log level in calling module to DEBUG to see log messages
+log = logging.getLogger(__name__)
+
+
+def makeADir(dirName, flagKeepDir=False):
+    """ Make directories """
+
+    # If directory already exists - Delete directory first is default
+    if os.path.isdir(dirName):
+        if flagKeepDir:
```
matthiasto, line 24: so you keep it this way where the default is remove dir?

awirb: What is your opinion, shall I put it the other way round and make a FlagRemDir?

avaframe/avaframeCfg.ini
awirb: Ok fine I remove it

avaframe/in3Utils/fileHandlerUtils.py
matthiasto: okay lets keep it like this

awirb: I think you are right, I changed it to flagRemDir

PR #87, opened 2020-08-28
Pretify plots

matthiasto:
adding plot settings in out3SimpPlot
changing plots in AB and Aimec
harmonizing file headers


avaframe/out3SimpPlot/plotSettings.py
```diff
@@ -0,0 +1,50 @@
...
+    This file is part of Avaframe.
+"""
+
+import seaborn as sns
+from matplotlib import cm
+import copy
+import matplotlib
+
+# define figure dimentions
+figW = 12
```
awirb, line 13: Can you make this a dictionary so that there is a option to chose 1, 2, 3, 4 column figure and 1, 2, 3, 4 rows?

avaframe/out3SimpPlot/plotSettings.py
```diff
@@ -0,0 +1,50 @@
...
+figH = 8
```
awirb, line 14: same here

avaframe/out3SimpPlot/plotSettings.py
```diff
@@ -0,0 +1,50 @@
...
+#  'ytick.left': True})
+
+# hell white/green to dark blue
+cmapGB = sns.cubehelix_palette(8, start=.5, rot=-.75, as_cmap=True)
+cmapGB.set_bad(color='k')
+# hell pink to dark purple
+cmapPP = sns.cubehelix_palette(8, as_cmap=True)
+cmapPP.set_bad(color='k')
+
+cmapjet = copy.copy(matplotlib.cm.jet)
```
awirb, line 38: why do we need Jet?

awirb (changes requested):

avaframe/out3SimpPlot/plotSettings.py
matthiasto: it is just defined in case we want to use it, bun not used

PR #89, opened 2020-08-28
Aimec speed

matthiasto:
add speed to aimec


awirb (changes requested):
* copy cmap to get rid of Warning
* in plot2 make right y axis green labels, add units on left y axis, add legend for the points/cross, add title
* in plot 3 add legend for point/cross, make figure extent so big that all markers are visible
suggestion:
* would be nice to add a description in the docu about the plots
* add debug figs to the ones that get saved

PR #93, opened 2020-09-01
add log tests

matthiasto:
log tests #47 
cfg test #46


avaframe/tests/test_cfgUtils.py
```diff
@@ -0,0 +1,17 @@
+"""Tests for module logUtilis"""
+from avaframe.in3Utils import cfgUtils
+from avaframe.tests import test_logUtils
+import os
+import configparser
+# import logging
```
awirb, line 6: if not used remove

avaframe/tests/test_cfgUtils.py
```diff
@@ -0,0 +1,17 @@
...
+# import logging.config
```
awirb, line 7: same here

PR #98, opened 2020-09-04
Test proj

awirb:


avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -28,3 +26,14 @@ aimecDir = Outputs/com1DFA/
 calcPressureLimit = 10
 # width of the avalanche path domain for computations in Aimec
 domainWidth = 600
```
matthiasto, line 28: This is not needed anymore. It is set in the AIMEC.ini file
You don't even need to write this .txt file anymore

avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -28,3 +26,14 @@ aimecDir = Outputs/com1DFA/
...
+
+
+[PARAMATERVAR]
+# True if parameter variation of MU option shall be included
+flagVarPar = False
+# parameter to be varied, current option: Mu (friction parameter)
+varPar = Mu
```
matthiasto, line 35: can we change the Mu to something like depth? will it work?

matthiasto (changes requested):

avaframe/com1DFA/com1DFACfg.ini
awirb: Ok, now there is no section Aimec anymore in the .ini

avaframe/com1DFA/com1DFACfg.ini
awirb: I added the option to change the release area thickness

PR #99, opened 2020-09-04
Update moduleAna3AIMEC.rst

jtfisch:
few changes and a comment..


jtfisch:
review of aimec description... btw. good JOB! and we could formulate the "P_CROSS" etc more generally... runout etc could be "velocity/depth" based or even "pressure/energy" this really depends on the politicly/guideline decisive values and wouldnt change any content...

PR #105, opened 2020-09-15
adding a savefig.format option

matthiasto:
adding a savefig.format option:
when saving a plot, use fig.savefig('theNameOfTheFigWithNoExtension') without specifying the extension.
The format is specified in plotsettings.py:
matplotlib.rcParams["savefig.format"] = 'png'


avaframe/ana3AIMEC/ana3AIMEC.py
```diff
@@ -642,7 +642,7 @@ def analyzeArea(rasterTransfo, resAnalysis, pLim, newRasters, cfgPath, cfgFlags)
             pathResult = cfgPath['pathResult']
             projectName = cfgPath['dirName']
             outname = ''.join([pathResult, os.path.sep, 'pics', os.path.sep,
-                               projectName, '_', str(i), '_compToRef', '.pdf'])
+                               projectName, '_', str(i), '_compToRef'])
```
fso42, line 645: This  (and below) see issue #107

avaframe/out3SimpPlot/plotSettings.py
```diff
@@ -14,7 +14,7 @@
 # define figure dimentions
 figW = 6
 figH = 6
-# define figure resolution
+# define figure resolution (dpi)
```
fso42, line 17: Should this and the safefig.format end up in the general .ini?
I think it would make sense to put it there..

fso42 (changes requested):

avaframe/out3SimpPlot/plotSettings.py
matthiasto: I could move it there yes.  But it is harder to all then no?

fso42 (approved):

PR #117, opened 2020-09-28
Add text

awirb:
* Add a text box with max, mean and min value in difference plot
* Add option to filter qick plot data using a parameter of the simulation name, e.g. Mu, simulation type, ...
* Add outputformart option for saving figures as parameter

### DIFF
--- a/avaframe/out3SimpPlot/outQuickPlot.py
+++ b/avaframe/out3SimpPlot/outQuickPlot.py
@@ -85,6 +85,16 @@ def quickPlot(avaDir, suffix, cfg, simName):
         ny_loc = int(nx * 0.5)
         nx_loc = int(ny * 0.5)
 
+        # Difference between datasets
+        dataDiff = data1 - data2
+        diffMax = np.amax(dataDiff)
+        diffMin = np.amin(dataDiff)
+        diffMean = np.mean(dataDiff)
+        
+        # Location of box
+        nybox = int(nx * 0.2)
+        nxbox = int(ny * 0.2)
+
         # Plot data
         # Figure 1 shows the result parameter data
         fig = plt.figure(figsize=(figW*3, figH))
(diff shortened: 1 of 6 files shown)


avaframe/runQuickPlot.py
```diff
@@ -24,11 +24,14 @@
 log.info('Current avalanche: %s', avalancheDir)
 
 # REQUIRED+++++++++++++++++++
+# Which parameter to filter data, e.g. varPar = 'simType', values = ['null'] or varPar = 'Mu', values = ['0.055', '0.155']
+ # values need to be given as list, also if only one value
 outputVariable = ['pfd', 'ppr']
-simName = 'entres'
+values = ['null']
+varPar = 'simType'
```
fso42, line 31: Would it make sense to include these in a ini file as default values?

fso42 (approved):

avaframe/out3SimpPlot/outQuickPlot.py
```diff
@@ -109,13 +119,15 @@ def quickPlot(avaDir, suffix, cfg, simName):
         ax2.set_title(title)
         ax3 = fig.add_subplot(133)
         cmap = cmapdiv
-        im3 = plt.imshow(data1-data2, cmap=cmap,
+        im3 = plt.imshow(dataDiff, cmap=cmap,
                          extent=[0, Lx, 0, Ly], origin='lower', aspect=nx/ny)
         fig.colorbar(im3, ax=ax3)
+        ax3.text(nybox, nxbox, 'Mean: %.2f\n Max: %.2f\n Min: %.2f' % (diffMean, diffMax, diffMin), bbox=dict(boxstyle="square", ec='white', fc='white'),
+        horizontalalignment='left', verticalalignment='bottom')
```
matthiasto, line 126: could we maybe add some units here? to be sure if its m or kPa or %

matthiasto (changes requested):
Could we add some doc to know what this does exactly and where to put the inputs, I am not sure I understood everything it does

avaframe/out3SimpPlot/outQuickPlot.py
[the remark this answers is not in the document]
awirb: the difference plot can be for pfd, ppr, pv, etc. and so far we also dont put units on the color bars

awirb: same as above, then i think we have to introduce the units for the colorbars too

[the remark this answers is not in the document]
matthiasto: okay but what appears on the 3rd plot is a value with the unit of what you look at? or some % ?
Just to be sure

avaframe/runQuickPlot.py
awirb: I added a outQuickPlotCfg.ini file with default values, and now the values and parameter used for filtering the data can be set there

avaframe/out3SimpPlot/outQuickPlot.py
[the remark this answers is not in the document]
awirb: I added the unit now

awirb: Added the unit

PR #118, opened 2020-09-28
Clean ainec visu

### LINKED ISSUE #115: Clean / Split / Simplify outAIMEC.py
There is a lot of code duplicity and repetitions in outAIMEC.py

We need to simplify it, split out smaller functions and make it more readable

matthiasto:
fix #115 
Cleaning the out3Aimec plots by creating a plot function
Updating the plots to use the new color map system
Sorting the input files properly before using them


avaframe/avaframeCfg.ini
```diff
@@ -6,6 +6,8 @@
 # Path to avalanche directory
 avalancheDir = data/avaHockeySmoothSmall
 
+# should the results be plotted with a continuous (True) or discrete colormap 
+continuousCmap = False
```
fso42, line 10: Please move this to Flags or move it to the plotting ini file ( I feel the second option is the better one)

avaframe/out3SimpPlot/plotSettings.py
```diff
@@ -8,11 +8,15 @@
...
 import matplotlib
+from matplotlib.image import NonUniformImage
 import cmocean
 import copy
 
+from avaframe.in3Utils import cfgUtils
 from avaframe.out3SimpPlot.makePalette import *
 
+cfgMain = cfgUtils.getGeneralConfig()
+continuousCmap = cfgMain['MAIN'].getboolean('continuousCmap')
```
fso42, line 20: This is not a good idea:
apart from it not working on a clean install (see the test build of this PR (pytest) ), 
it is dangerous to read a config twice-> what if the user decides to dynamically change a setting variable in his main script without being aware that a module reads it again...
-> we should keep this the way it is at the moment: the main script reads the cfg and supplies this to the modules...

avaframe/out3SimpPlot/plotSettings.py
```diff
@@ -37,7 +41,7 @@
 matplotlib.rcParams["savefig.format"] = 'png'
```
fso42, line 41: Stuff like this should go to an ini file

avaframe/out3SimpPlot/plotSettings.py
```diff
@@ -131,4 +133,20 @@
...
+###################################
+def myNonUnifIm(ax, x, y, z, xlab, ylab, **kwargs):
+    im = NonUniformImage(ax, **kwargs)
+    im.set_data(x, y, z)
+    ref = ax.images.append(im)
+    ax.set_xlim([x.min(), x.max()])
+    ax.set_ylim([y.min(), y.max()])
+    ax.set_xlabel(xlab)
+    ax.set_ylabel(ylab)
+    return ref, im
```
fso42, line 152: As mentioned in main comment -> extra .py file

fso42 (changes requested):
The logic of plotSettings.py needs to change.  

- We need an .ini file with the pure settings part. 

- There are quite some parameter like (figW, figH, lw, ms, fs etc) which should go to this ini file

- And a new .py file with takes the ini and does the setting logic for matplotlib (eg. plotUtils.py)
 -> can also contain those general function like the myNonUnifIM

The build is currently failing (build pytest) -> I added a note at the beginnig of plotSettings.py

I think we need to address these issues first, so I did not test the plots yet, and also did not look at the outAIMEC yet... 

Also be aware: once this is PR is merged, out1 and out3 are going to be renamed like discussed in our meeting (see Admin techDoc) -> out1Peak and out3Plot

matthiasto:
The plotting settings are now given in the plotUtils.ini in in3Utils. They are read by plotUtils.py still in in3Utils. The plotting environment is also set there.

fso42 (approved):

PR #123, opened 2020-10-01
Dfa doc

matthiasto:
- Working on the dfa appendix doc (correcting typos and detailing the numerics part)
- Adding sphinxcontrib.bibtex for managing bibliography (this means we can directly use the .bib from latex) use :cite:`Zw2000,ZwKlSa2003` to refer to some articles. The biblio now only appears in the section references at the end of the documentation. Maybe it could be nice to have a local one at the end of the page the user is reading.


avaframe/out3Plot/outAIMEC.py
```diff
@@ -386,7 +386,7 @@ def resultVisu(cfgPath, cfgFlags, rasterTransfo, resAnalysis, plim):
     # Get input data
     fnames = cfgPath['pressurefileList']
 
-    flag = cfgFlags['typeFlag']
+    flag = float(cfgFlags['typeFlag'])
```
fso42, line 389: Why the float here? string or int should be fine, so no conversion is necessary, right?

requirements.txt
```diff
@@ -4,3 +4,4 @@ pyshp
 scipy
 cmocean
 seaborn
+sphinxcontrib-bibtex
```
fso42, line 7: Could this be moved to docs/requirements.txt? 
To keep this requirements as clean as possible. A user who does not build the documentation, wanting only to run the model, does not need this module...

avaframe/out3Plot/outAIMEC.py
fso42: Ah, i forgot about the missing type conversion of the ini system

.github/workflows/runTestSinglePython.yml
```diff
@@ -26,6 +26,7 @@ jobs:
         pip install flake8 pytest
         pip install pytest-cov
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        if [ -f docs/requirements.txt ]; then pip install -r docs/requirements.txt; fi
```
fso42, line 29: Is this really needed? As the test is not building the docs... and readthedocs knows about the docs/req...  anyways...

matthiasto: I havent tried without but if you say reads the docs knows about it, I remove it

fso42: If readthedocs does not automatically find it, we can point to it in the Advanced settings of the avaframe project on rtd...

PR #126, opened 2020-10-02
Addreportfunction

awirb:
* Add generate markdown report
* Add function to plot all peak fields
* Add function to create one specific peak file
* Add docu for out3Plot, out3SimpPlot, log2Report

PR #128, opened 2020-10-02
Check local ini

matthiasto:
Modified the cfgUtils to check if local.ini matches the remote one and corrects it if necessary.
To see how it works, run python runTmp1Ex.py


fso42 (changes requested):
Don't know exactly what went wrong, but this PR has all the changes from PR #126 included. Please reduce this PR to just include your changes.

matthiasto:
Created another branch with just the check local config file changes

PR #129, opened 2020-10-05
Fix #124 checking if local ini up to date

### LINKED ISSUE #124: check for new cfg .ini settings
The configuration is read from .ini file. The local_ one if it exists, the remote one if local not there.
Problem is that if the remote changed without the user being aware of it (for example change of name for a variable or new variables), then you will for sure get errors and not necessarily  know where it comes from.
I suggest we check for differences between the remote and loacal cfg entries warn the user about such differences and take the remote entries as reference (if an entry is missing in the local take the value of the missing entry from the remote .ini)

matthiasto:
Modified the cfgUtils to check if local.ini matches the remote one and corrects it if necessary.
To see how it works, run python runTmp1Ex.py

### DIFF
--- a/avaframe/tests/test_tmp1Ex.py
+++ b/avaframe/tests/test_tmp1Ex.py
@@ -11,5 +11,4 @@ def test_tmp1ExMain(capfd):
 
     tmp1Ex.tmp1ExMain(cfg)
     out, err = capfd.readouterr()
-    print(out)
     assert out == 'In tmp1Example\n'
--- a/avaframe/tests/test_tmp1Ex.py
+++ b/avaframe/tests/test_tmp1Ex.py
@@ -11,4 +11,5 @@ def test_tmp1ExMain(capfd):
 
     tmp1Ex.tmp1ExMain(cfg)
     out, err = capfd.readouterr()
+    print(out)
     assert out == 'In tmp1Example\n'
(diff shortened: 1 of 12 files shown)


fso42 (changes requested):
I think the compareConfig function is such an essential function to get right, it should have pytests that check it thoroughly to e.g. make sure:

- the correct values are picked i.e. fileoveride -> local -> defaultCfg
- both for sections, and variables
- for complete sections, partial sections, new and old sections
If you feel this is too much for this PR, open an issue and merge this...

Another question we need to address: 
how do we handle the situation in which we change the default value of a variable e.g. X, to make sure users (who just copied the default to local to change unrelated variable A ) are aware of the change?

avaframe/tests/test_tmp1Ex.py
```diff
@@ -11,4 +11,5 @@ def test_tmp1ExMain(capfd):
 
     tmp1Ex.tmp1ExMain(cfg)
     out, err = capfd.readouterr()
+    print(out)
```
fso42, line 14: This is a debug statement, right?

matthiasto: oups forgot that... is it possible to get the out from a log.info or do we have to use print() (see the tmp1ExMain)

PR #130, opened 2020-10-05
fixAB

matthiasto:
Now taking into account the sign of the angle of the slope.
Before
angle = [40, 35, 30, 25, 20, 15, 10, 0, 10, 15, 20, 25]
no beta point found (under 10° with the 30 following m under 10°, here only 20m)
Now 
angle = [40, 35, 30, 25, 20, 15, 10, 0, -10, -15, -20, -25]

### DIFF
--- a/avaframe/in2Trans/geoTrans.py
+++ b/avaframe/in2Trans/geoTrans.py
@@ -281,7 +281,7 @@ def prepareAngleProfile(beta, AvaProfile):
         log.warning('No split Point given!')
         CuSplit = 0
     ds = np.abs(s - np.roll(s, 1))
-    dz = np.abs(z - np.roll(z, 1))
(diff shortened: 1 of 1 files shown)


avaframe/in2Trans/geoTrans.py
```diff
@@ -281,7 +281,7 @@ def prepareAngleProfile(beta, AvaProfile):
         log.warning('No split Point given!')
         CuSplit = 0
     ds = np.abs(s - np.roll(s, 1))
-    dz = np.abs(z - np.roll(z, 1))
+    dz = (np.roll(z, 1) - z)
```
fso42, line 284:
```suggestion
    dz = np.roll(z, 1) - z
```

fso42 (changes requested):

PR #135, opened 2020-10-06
Update documentation

awirb:
* Update documentation

### DIFF
--- a/docs/moduleCom1DFA.rst
+++ b/docs/moduleCom1DFA.rst
@@ -51,7 +51,7 @@ The simulation results are saved to: *Outputs/com1DFA* and include:
 To run
 --------
 
-* create an avalanche directory with required input files
+* create an avalanche directory with required input files - for this task you can use runInitializeProject.py
 * copy com1DFACfg.ini to local_com1DFACfg.ini and set your desired settings
 * run runCom1DFA.py
 
--- a/docs/moduleCom1DFA.rst
+++ b/docs/moduleCom1DFA.rst
@@ -16,7 +16,7 @@ Input
 ---------
 
 The module requires an avalanche directory, that follows a specified folder structure.
-This avalanche directory can be created by running: runInitializeProject.py.
+This avalanche directory can be created by running: ``runInitializeProject.py``.
 In the directory Inputs, the following files are required:
 
 * digital elevation model as .asc file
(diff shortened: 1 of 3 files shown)


docs/moduleCom1DFA.rst
```diff
@@ -51,7 +51,7 @@ The simulation results are saved to: *Outputs/com1DFA* and include:
 To run
 --------
 
-* create an avalanche directory with required input files
+* create an avalanche directory with required input files - for this task you can use runInitializeProject.py
```
matthiasto, line 54: maybe add the link to the page

docs/moduleIn3Utils.rst
```diff
@@ -194,3 +194,32 @@ header information as well as the raster data in a numpy array in a dictionary.
...
+
+
+Input
+-------
+
+* path to NameOfAvalanche
+
+This path is specified in the configuration file *avaframeCfg.ini* by setting the parameter *avalancheDir*.
+
+Output
```
matthiasto, line 222: Could we maybe add an option in this function to make a test example so like:
avaTutorial
python runInitializeProject.py

and this would generate a avaTutorial folder structure with already the rel, res ... files, so ready to run.

docs/moduleIn3Utils.rst
```diff
@@ -194,3 +194,32 @@ header information as well as the raster data in a numpy array in a dictionary.
...
+-------
+
+* NameOfAvalanche directory
```
matthiasto, line 225: maybe add a section To run like in all other modules

docs/moduleIn3Utils.rst
```diff
@@ -194,3 +194,32 @@ header information as well as the raster data in a numpy array in a dictionary.
...
+		Outputs/
+		Work/
+
+
+Input
+-------
+
+* path to NameOfAvalanche
+
+This path is specified in the configuration file *avaframeCfg.ini* by setting the parameter *avalancheDir*.
```
matthiasto, line 220: how will the user get the avaframeCfg.ini in the first place?

matthiasto (changes requested):

docs/moduleIn3Utils.rst
awirb: you mean that also fetches a DEM and release area scenario? there are the standard tests that are ready to run though..

docs/moduleIn3Utils.rst
awirb: Done

docs/moduleCom1DFA.rst
awirb: thanks for the hint, included a link

docs/moduleIn3Utils.rst
awirb: avaframeCfg.ini is found in /avaframe, I added in the section to run, that the avaframeCfg.ini first should be copied to local and then avalancheDir needs to set

docs/moduleIn3Utils.rst
awirb: the standardtests are already fully accessible

PR #156, opened 2020-10-16
Update standard tests

awirb:
* Add Warning in report if min difference exceeds threshold 
* add standardTests cfg file


avaframe/com1DFA/standardTests_com1DFACfg.ini
```diff
@@ -0,0 +1,48 @@
+### Config File - This file contains the main settings for the simulation run
+## Copy to  local_com1DFACfg.ini and set you parameters
+# This file is part of Avaframe.
+
+
+[GENERAL]
+# Path to samos executable and AK_Attributes files
+samosAT = /home/anna/repos/gitea/AvaFrame_SamosDFA/samosAT -files /home/anna/repos/gitea/AvaFrame_SamosDFA/files/AK_Attributes/
```
fso42, line 8: Please remove your local path :-)

avaframe/log2Report/generateCompareReportCfg.ini
```diff
@@ -2,6 +2,9 @@
 ## Copy to  local_outQuickPlotCfg.ini and set you parameters
 # This file is part of Avaframe.
 
+[GENERAL]
+# limit to produce warning for difference in peak fields
+diffLim = 0.00001
```
fso42, line 7: Should we maybe use percentage instead of a fixed value?

fso42 (changes requested):
I like your idea of having a seperate .ini file. I just wonder if we could take it a step further and dump the current benchmark ini into each of our test cases and load it for every case seperately?

avaframe/com1DFA/standardTests_com1DFACfg.ini
awirb: Done

avaframe/log2Report/generateCompareReportCfg.ini
awirb: Good idea! What percentage of the min difference would be suitable?

fso42: Mh, that's up for discussion. How about we start with 0.01 % ?

awirb: fine!

awirb:
* now a avaName_com1DFACfg.ini file for each benchmark located in the benchmark folders
* samosAT executable path still read form com1DFACfg.ini
* Warning if difference is diffLim percent of max value of benchmark result

fso42 (approved):

PR #159, opened 2020-10-19
Rename com1DFA; clean directory helper

fso42:
**BREAKING CHANGE**:
- recompilation of com1DFA exe is highly recommended
- adjust your local ini files for com1DFA

Regarding version 0.1: 
- add a runOperational file: this is our algorithm target for version 0.1 (please look into parts you can provide) @matthiasto: I'd like to have the call to alphaBeta as simple as shown in this file. Does a function like this already exist?

- Helper to clean avalanche directories from *Outputs*, *Work* and *.log files

### DIFF
--- a/avaframe/com1DFA/com1DFACfg.ini
+++ b/avaframe/com1DFA/com1DFACfg.ini
@@ -4,8 +4,8 @@
 
 
 [GENERAL]
-# Path to samos executable and AK_Attributes files
-samosAT = ./samosAT -files files/AK_Attributes/
+# Path to com1DFA executable and AK_Attributes files
+com1Exe = ./com1DFA -files files/AK_Attributes/
 # Name of run cint file template
 RunCint = runBasic
 # Input format of release are files
--- a/avaframe/com1DFA/com1DFACfg.ini
+++ b/avaframe/com1DFA/com1DFACfg.ini
@@ -5,7 +5,7 @@
 
 [GENERAL]
 # Path to com1DFA executable and AK_Attributes files
-com1Exe = ./com1DFA -files files/AK_Attributes/
+com1Exe = ./com1DFA.exe -files files/AK_Attributes/
 # Name of run cint file template
 RunCint = runBasic
 # Input format of release are files
--- a/avaframe/in3Utils/initializeProject.py
+++ b/avaframe/in3Utils/initializeProject.py
@@ -26,7 +26,7 @@ def cleanSingleAvaDir(avaDir, keep=None, deleteOutput=True):
     Expects a avalanche directory name as string
     and optional:
     a log name to keep (and not delete)
-    Boolean to be able to avoid deletion of Outputs
+    Boolean to be able to avoid deletion of Outputs (true by default)
     '''
 
     # check for empty avaDir name, abort if empty
--- a/avaframe/in3Utils/initializeProject.py
+++ b/avaframe/in3Utils/initializeProject.py
@@ -26,13 +26,13 @@ def cleanSingleAvaDir(avaDir):
 
     # check for empty avaDir name, abort if empty
     if not avaDir:
-        log.warning("AvaDir variable is undefined, getting the f*** out of here")
+        log.warning("AvaDir variable is undefined, returning. ")
         return('AvaDir is empty')
 
     # make sure avaDir is a string
     isString = isinstance(avaDir, str)
     if not isString:
-        log.warning("AvaDir is not a string, getting the f*** out of here")
+        log.warning("AvaDir is not a string, returning")
         return('AvaDir is NOT a string')
 
     # Info to user
--- a/avaframe/in3Utils/initializeProject.py
+++ b/avaframe/in3Utils/initializeProject.py
@@ -6,8 +6,56 @@
 
 import os
 import logging
+import shutil
 
 log = logging.getLogger(__name__)
+
+def _checkForFolderAndDelete(baseDir, folderName):
+    '''Helper function ONLY USE WITH CHECKED INPUT'''
+    fPath = os.path.join(baseDir, folderName)
+    try:
+        shutil.rmtree(fPath)
+    except FileNotFoundError:
+        log.debug("No %s folder found.", folderName)
+        pass
+
+def cleanSingleAvaDir(avaDir):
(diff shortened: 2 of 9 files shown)


avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -4,8 +4,8 @@
 
 
 [GENERAL]
-# Path to samos executable and AK_Attributes files
-samosAT = ./samosAT -files files/AK_Attributes/
+# Path to com1DFA executable and AK_Attributes files
+com1Exe = ./com1DFA -files files/AK_Attributes/
```
matthiasto, line 8: I tried to rename the exe com1DFA but since there is already a com1DFA directory, this is not possible.
I had to name it com1DFE.exe for it to work.

avaframe/in3Utils/initializeProject.py
```diff
@@ -6,9 +6,57 @@
...
+        pass
+
+def cleanSingleAvaDir(avaDir):
+    '''
+    Clean a single avalanche directory of the work and output directories
+    '''
+
+    # check for empty avaDir name, abort if empty
+    if not avaDir:
+        log.warning("AvaDir variable is undefined, getting the f*** out of here")
```
matthiasto, line 29: I would prefer not to use this type of language.

avaframe/in3Utils/initializeProject.py
```diff
@@ -6,9 +6,57 @@
...
+        return('AvaDir is empty')
+
+    # make sure avaDir is a string
+    isString = isinstance(avaDir, str)
+    if not isString:
+        log.warning("AvaDir is not a string, getting the f*** out of here")
```
matthiasto, line 35: same here

avaframe/runCom1DFA.py
```diff
@@ -33,7 +33,7 @@
 
 startTime = time.time()
 # Run Standalone DFA
-reportDictList = com1DFA.runSamos(cfg, avalancheDir)
+reportDictList = com1DFA.runCom1DFA(cfg, avalancheDir)
```
matthiasto, line 36: This is very confusing. There is a runCom1DFA.py and a runCom1DFA function in com1DFA/com1DFA.py. I got a bit lost trying to figure out what was happening.

avaframe/runOperational.py
```diff
@@ -0,0 +1,92 @@
...
+from avaframe.out3SimpPlot import outPlotAllPeak as oP
+from avaframe.in3Utils import cfgUtils
+from avaframe.in3Utils import logUtils
+from avaframe.in3Utils import initializeProject as initProj
+
+# Time the whole shebang
+startTime = time.time()
+
+# log file name; leave empty to use default runLog.log
+logName = 'runOperational'
```
matthiasto, line 23: I can't find the log that is created... is it supposed to be in the ava directory?

avaframe/runOperational.py
```diff
@@ -0,0 +1,92 @@
...
+# Load input parameters from configuration files
+
+# ----------------
+# Clean input directory(ies) of old work and output files
+initProj.cleanSingleAvaDir(avalancheDir)
+
+# ----------------
+# Run dense flow
+# cfg = cfgUtils.getModuleConfig(com1DFA)
+# reportDictList = com1DFA.runCom1DFA(cfg, avalancheDir)
```
matthiasto, line 48: This still create the old report. Is it supposed to?

matthiasto (changes requested):

avaframe/runOperational.py
fso42: Ah, yes, this needs to be caught: it is currently deleted in the clean script ...

avaframe/runOperational.py
fso42: What do you mean by old report?

matthiasto: the release1HX_null_dfa_0.155.html for example

avaframe/runStandardTests.py
```diff
@@ -50,8 +50,8 @@
 for avaDir in standardNames:
 
     # get path to executable
-    cfgSamosAT = cfgUtils.getModuleConfig(com1DFA)
-    samosAT = cfgSamosAT['GENERAL']['samosAT']
+    cfgCom1DFA = cfgUtils.getModuleConfig(com1DFA)
+    com1Exe = cfgCom1DFA['GENERAL']['com1Exe']
```
awirb, line 55: Question: should the avalancheDir be cleaned with cleanSingleAvaDir?

awirb (commented):
Idea: should we have an option in the cleanSingleAvaDir that only the folders for individual modules like com1DFA should be cleaned but not the plots?

avaframe/runStandardTests.py
fso42: yes. cleanSingleAva needs directory in which Ouputs , Work etc (potentially) live...

awirb: Would also be useful to add this here in the runStandardTests.py

fso42: yes, definitely

avaframe/runOperational.py
fso42: ah, @awirb correct me if I'm wrong, this comes from the generate export function of com1DFA itself, right? We can remove this, but that's not urgent... (i.e. version 0.2)

awirb: Correct, these reports are created by the generate_report function in the run cint files of com1DFA. This function also exports the peak files.. and at the moment include more parameters than the .md report

fso42:
> Idea: should we have an option in the cleanSingleAvaDir that only the folders for individual modules like com1DFA should be cleaned but not the plots?

You mean so we can also clean at the end?

awirb:
No but for example there is in Outputs a folder out3SimpPlot and we don't want to get rid of this one.. but actually maybe it is better to just clean all the stuff in Outputs so it doesn't become confusing which data belongs to which simulation..

fso42:
I'll include a function that just cleans the Work directory, but leaves the ouputs untouched. This way we could clean at the end and save space...

matthiasto (approved):

PR #168, opened 2020-10-22
check inputs

matthiasto:
Check that max 1 entrainment and resistance shp file is given in input

### DIFF
--- a/avaframe/ana3AIMEC/ana3AIMEC.py
+++ b/avaframe/ana3AIMEC/ana3AIMEC.py
@@ -46,7 +46,7 @@ def readAIMECinputs(avalancheDir, dirName='com1DFA'):
     profileLayer = glob.glob(os.path.join(avalancheDir, 'Inputs', 'LINES',
                                           '*aimec*.shp'))
     try:
-        message = 'There should be exactly one path_aimec.shp file containing the avalanche paths in ' + avalancheDir + '/Inputs/LINES/'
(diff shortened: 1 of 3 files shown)


avaframe/ana3AIMEC/ana3AIMEC.py
```diff
@@ -45,10 +45,20 @@ def readAIMECinputs(avalancheDir, dirName='com1DFA'):
 
     profileLayer = glob.glob(os.path.join(avalancheDir, 'Inputs', 'LINES',
                                           '*aimec*.shp'))
+    try:
+        message = 'There should be exactly one path_aimec.shp file containing the avalanche paths in ' + avalancheDir + '/Inputs/LINES/'
```
fso42, line 49:
```suggestion
        message = 'There should be exactly one path_aimec.shp file containing (multiple) avalanche paths in ' + avalancheDir + '/Inputs/LINES/'
```
Maybe? To make put more emphasis on that you can have multiple paths in one shape file?

fso42 (approved):
Just as small 'maybe' comment...

avaframe/ana3AIMEC/ana3AIMEC.py
matthiasto: for aimec there should only be one path. For AB I agree that there can be multiple paths in the shp file

fso42: ah ok, then `paths -> path`

PR #172, opened 2020-10-28
Updating documentation

matthiasto:
- adding intro (basically a copy of the one in avafrme.org)
- adding details to the develop section (how to install dependencies...). We need to test if this workflow (installation and running the routine from outside Avaframe works on every linux distribution)
- adding a get started example (a bit as a tutorial)
- moving the asci and shp file routines to in2Trans and all other tools to in3Utils

You can have a look at the hidden doc [here](https://docs.avaframe.org/en/runopedoc/index.html)


awirb:
I suggest to make a separate Chapter: Installation and move how to install from Develop Avaframe to there

docs/develop.rst
```diff
@@ -13,17 +11,44 @@ Clone repository::
...
+
+    conda create --name avaframe_env
+
+Don't forget to activate it before going on::
+
+    conda activate avaframe_env
+
+Move to ``Avaframe/`` repository and install the different modules::
+
+    conda install --name avaframe_env --file avaframe_env_spec.txt
```
awirb, line 36: I think installation of make is missing

docs/develop.rst
```diff
@@ -13,17 +11,44 @@ Clone repository::
...
+
+Or the equivalent pip commands.
+
+Finally install avaframe::
+
+  conda develop .
```
awirb, line 42: conda develop, conda-build missing?

awirb (changes requested):
Specific text edits in next commit

docs/getStarted.rst
```diff
@@ -0,0 +1,177 @@
...
+================
+
+Even-though most of the routines in Avaframe are automatized, running them still
+requires a minimum amount of python scripting.
+Here we provide an example on how to perform simulations with Avaframe,
+but let's start with some information on the utilities to create logs and set the desired configurations.
+
+Utilities
+----------
+
```
fso42, line 11: This is fine, but I suggest we have a 'in 5 minutes' (or similiarly named) section at the beginning, in which we have an enumerated list with the absolute bare minimum command and only the absolute necessary explanations to run a simulations. I.e.: 

- Clone the DFA repo by git clone...
- Clone repository by git clone...
- Change to directory i.e cd AvaFrame
- Local install by pip install ...
- Edit avaframe com2DFACfg.ini to 

and so on

docs/installation.rst
```diff
@@ -0,0 +1,71 @@
...
+
+* numpy
+* matplotlib
+* pyshp
+* scipy
+* cmocean
+* seaborn
+* pytest
+
+If you are working on Linux, we suggest to work with anaconda (which you can install following `these instructions <https://docs.anaconda.com/anaconda/install/linux/>`_)
```
fso42, line 23: mention miniconda as well (which I would suggest as default)

docs/installation.rst
```diff
@@ -0,0 +1,71 @@
...
+
+    conda create --name avaframe_env
+
+Don't forget to activate it before going on::
+
+    conda activate avaframe_env
+
+Move to ``Avaframe/`` repository and install the different modules by using the provided installation setup file::
+
+    conda install --name avaframe_env --file avaframe_env_spec.txt
```
fso42, line 36: see main comment

docs/installation.rst
```diff
@@ -0,0 +1,71 @@
...
+
+In order to build the documentation you need to install make ::
+
+  sudo apt install make
+
+Then go to the ``docs\`` directory and run::
+
+  make html
+
+Html files of the documentation can be found in the _build directory.
```
fso42, line 71: I think this should stay in the develop sections. Only someone developing will most likely work on this, all others can look at the online docs...

docs/moduleConfiguration.rst
```diff
@@ -0,0 +1,24 @@
...
+
+The ``getModuleConfig`` function reads the settings from the configuration file (``tmpEx.ini``
+in our example) and writes these settings to the log file. The default settings can be found in the
+configuration file provided within each module.
+It is possible to modify these settings by creating a copy of the configuration
+file (``local_`` followed by the name of the original configuration file) and set the
+desired values of the individual parameters. If there is a configuration file with the
+prefix ``local_``, the ``getModuleConfig`` function reads the settings from this file.
+Another option is to read the settings from a different configuration file, therefore the path to this file
+has to be passed to the ``getModuleConfig`` function directly.
```
fso42, line 24: The order of ini files should be made clearer, i.e. Override -> local_... -> default.ini

fso42 (changes requested):
- I think we should not use a file like avaframe_env_spec unless we really absolutely have to! This requests explicit versions, and I think we should avoid any version setting unless we run into problems. Furthermore this requires us to update requirements and this file and keep them in sync. Apparently conda also allows for installation from a requirements.txt file, we should try this first. 

> conda install --file requirements.txt
> or here
> https://www.technologyscout.net/2017/11/how-to-install-dependencies-from-a-requirements-txt-file-with-conda/
> 

- We need to note which Linux versions this was tested on

docs/installation.rst
awirb: moved to develop.rst

docs/getStarted.rst
awirb: Included

fso42 (approved):

awirb (approved):

Issue #177, opened 2020-11-05
Make log less 'chatty'


fso42:
I.e. move a lot of log feedback from Info to debug

Currently it is hard to follow what is going on...

PR #181, opened 2020-11-09
Update log levels used in com1DFA, plot functions and fileHandlerUtils 

awirb:
* contribute to issue #177

### DIFF
--- a/avaframe/out3Plot/outQuickPlot.py
+++ b/avaframe/out3Plot/outQuickPlot.py
@@ -121,7 +121,7 @@ def generatePlot(dataDict, avaName, outDir, cfg, plotDict):
     ax[1].legend()
     fig.savefig(os.path.join(outDir, 'Profiles_%s_%s.%s' % (avaName, simName, outputFormat)))
 
-    log.info('Figures saved to: %s' % outDir)
+    log.debug('Figures saved to: %s' % outDir)
 
     if cfg['FLAGS'].getboolean('showPlot'):
         plt.show()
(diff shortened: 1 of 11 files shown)


avaframe/out3Plot/outQuickPlot.py
```diff
@@ -121,7 +121,7 @@ def generatePlot(dataDict, avaName, outDir, cfg, plotDict):
     ax[1].legend()
     fig.savefig(os.path.join(outDir, 'Profiles_%s_%s.%s' % (avaName, simName, outputFormat)))
 
-    log.info('Figures saved to: %s' % outDir)
+    log.debug('Figures saved to: %s' % outDir)
```
fso42, line 124: mh, maybe leave this on level info? This is something the user might look for quite often...

avaframe/runCom1DFA.py
```diff
@@ -24,8 +24,8 @@
 
 # Start logging
 log = logUtils.initiateLogger(avalancheDir, logName)
-log.info('MAIN SCRIPT')
-log.info('Current avalanche: %s', avalancheDir)
+log.debug('MAIN SCRIPT')
+log.debug('Current avalanche: %s', avalancheDir)
```
fso42, line 28: I think these should stay in Info, as this is an important info...

fso42 (changes requested):

matthiasto (commented):
Looks good, I agree with Felixs comments.

avaframe/out3Plot/outQuickPlot.py
awirb: Changed to info level

avaframe/runCom1DFA.py
awirb: After loading the configuration in com1DFA there is an info about 'Your current avalanche...' so this is why I removed this here, but changed it back to be info

fso42: ahhh, sorry, forgot about that one. Then keep the debug level (since config *has* to be read anyway)

awirb: ok I have set it back to debug :)

fso42 (approved):

PR #190, opened 2020-11-13
read inputs with function and time stepping option

awirb:
* Add getInputData function to read dem files, release areas,...
* Add timeDiscretizations file where we can add different options for computing time steps, using time different time discretization schemes, etc. ...

* Restructure reading of inputs for com1DFA 
* add module in1Data with function for reading input data files 
* move generation of work and output directory to in3Utils


avaframe/com1DFAPy/timeDiscretizations.py
```diff
@@ -0,0 +1,66 @@
...
+#
+#
+#     Output:
+#
+#     """
+#
+#
+#
+#     #
+#     return
```
fso42, line 66: If not needed, please remove

avaframe/com1DFAPy/utilities.py
```diff
@@ -0,0 +1,93 @@
...
+log = logging.getLogger(__name__)
+
+
+def getInputData(avaDir, cfg):
+    """ Fetch input datasets required for simulation
+
+    Input:
+            Directory
+
+    Output:
```
fso42, line 31: Why not use a general function? Or provide this function generally? It seems to be a general enough task...

fso42 (changes requested):

avaframe/com1DFAPy/timeDiscretizations.py
awirb: Oh yes not needed, removed

matthiasto (approved):

avaframe/com1DFAPy/utilities.py
awirb: moved to in1Data

avaframe/com1DFAPy/timeDiscretizations.py
```diff
@@ -0,0 +1,50 @@
...
+from avaframe.in3Utils import cfgUtils
+from avaframe.in3Utils import logUtils
+
+
+# create local logger
+# change log level in calling module to DEBUG to see log messages
+log = logging.getLogger(__name__)
+
+
+def getTimeStep(particles, dem, cfg):
```
fso42, line 25: Maybe make it more desciptive by naming int getTimeStepCFL or something similar?

avaframe/com1DFAPy/timeDiscretizations.py
```diff
@@ -0,0 +1,50 @@
...
+    # courant number
+    cMax = float(cfg['cMax'])
+
+    # compute stable time step
+    # if velocity is zero - divided by zero error so to avoid:
+    if vmax == 0.0:
+        dtStable = float(cfg['maxdT'])
+    else:
+        dtStable = (cMax * csz) / vmax
+
```
fso42, line 44: So this could lead to an extremely small timestep which would basically grind the computations to a halt. Should we implement some checking, like if the timestep is below a certain limit, we try with a less conservative courant number up to a certain point?
Or does this make it way to overcomplicated for now?

fso42 (changes requested):

avaframe/com1DFAPy/timeDiscretizations.py
awirb: The way it is now if the velocity is zero and hence no time step can be computed according to the function, it depends on what you set in the .ini file as maxdT. Which is where at the moment the time step that is used is read. I would implement the suggestion in the future in case we need the cfl time step for the computations.

avaframe/com1DFAPy/timeDiscretizations.py
awirb: done

fso42 (approved):

PR #192, opened 2020-11-16
Add export of peak files

awirb:
* add export of peakfields to Outputs/com1DFAPy/peakFiles as .asc files
* add clean directories 
* add out1Peak plot all peak files to Outputs/com1DFAPy/reports


fso42 (approved):
I think we should rename com1DFAtoolsGfg.init to com1DFACfg.ini right away.

PR #221, opened 2020-12-01
add asc DEM

matthiasto:
Here are the DEM .asc files for real avalanches.
The cellsize is not 5m, I do not know what I should do regarding this point
I also added the bash script to extract the DEM from a shape file of the extend desired and the geotiff data


fso42:
I suggest we force 3 DEM to have the perfect 5m value, and leave the others as is. Our framework should be able to handle both...

fso42:
Maybe add an tinyHelper directory in the root and move the sh script there?

PR #228, opened 2020-12-07
adding AB and Aimec Path

matthiasto:


fso42 (changes requested):
I adjusted some paths. 
Two issues remain:

- Do we want the AIMEC path to always reflect the AB path? I adjusted some (where there was a problem with the AB - shape files. 
- There seems to be a problem with the DGM files, visible when you calculate slope or set the style to hillshade. A grid pattern emerges (see attached screen shots. 

Hillshade of Hit
Screenshot_2020-12-07_19-27-12

Slope of Gar
Screenshot_2020-12-07_19-13-38

PR #244, opened 2020-12-17
Add stats functions and rename ana4Prob to ana4Stats

### LINKED ISSUE #239: Rename ana4Prob to ana4Stats
To show that ana4 is not only for probabilities, but also for other statistical analyses...

awirb:
* adjust all scripts to use ana4Stats instead of ana4Prob fix #239 
*  update pytests 
* add continuous version of aimec colormap to plotUtils 
* add names for 'ppr', 'pfd', 'pv' in plotUtils that can be used for plot labelling 
* rename release area scenarios-> remove underscores


avaframe/out3Plot/plotUtilsCfg.ini
```diff
@@ -41,3 +41,9 @@ contCmap = False
 unitppr = kPa
 unitpfd = m
 unitpv = $ms^{-1}$
+unitrelTh = m
```
fso42, line 44: Should we adjust all these variables names to camelCase?
ie unitRelTh, namePpr or namePPR? not sure....

fso42 (approved):

avaframe/out3Plot/plotUtilsCfg.ini
awirb: No this won't work as they are build inside the scripts using the resultType variable that we use throughout e.g. in the simulation results naming and these area 'ppr', 'pfd' and 'pv'

PR #246, opened 2020-12-21
update compute dist

awirb:
* update naming of plots, logs 
* add write and read cfg function saved to Outputs

### DIFF
--- a/avaframe/in1Data/computeFromDistributionCfg.ini
+++ b/avaframe/in1Data/computeFromDistributionCfg.ini
@@ -17,5 +17,5 @@ support = 10000
 name = scenario70
 # if min max value shall be included true, default False
 flagMinMax = False
-# name of distribution
+# name of distribution i.e. pert [Pert], uniform [Uni]
 distType = Pert
--- a/avaframe/in1Data/computeFromDistributionCfg.ini
+++ b/avaframe/in1Data/computeFromDistributionCfg.ini
@@ -1,6 +1,6 @@
 ### Config File - This file contains the main settings for the computeFromDistributions
 ## Copy to  local_computeFromDistributionCfg.ini and set you parameters
-
+# This file is part of Avaframe.
 
 [GENERAL]
 # minimum (a) of desired distribution
@@ -17,3 +17,5 @@ support = 10000
 name = scenario70
 # if min max value shall be included true, default False
 flagMinMax = False
+# name of distribution
+distType = Pert
--- a/avaframe/in3Utils/cfgUtils.py
+++ b/avaframe/in3Utils/cfgUtils.py
@@ -162,17 +162,10 @@ def compareConfig(iniFile, modName, compare):
     return cfg
 
 
-def writeCfgTxt(avaDir, module, cfg):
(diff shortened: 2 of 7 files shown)


avaframe/in1Data/computeFromDistributionCfg.ini
```diff
@@ -17,3 +17,5 @@ support = 10000
 name = scenario70
 # if min max value shall be included true, default False
 flagMinMax = False
+# name of distribution
```
fso42, line 20: please list the possible options in the comment ie. [Pert, Uniform...] or whatever

avaframe/in3Utils/cfgUtils.py
```diff
@@ -159,3 +160,38 @@ def compareConfig(iniFile, modName, compare):
...
+    # set outputs
+    outDir = os.path.join(avaDir, 'Outputs')
+
+    # write to file
+    with open(os.path.join(outDir, '%s%s_settings.ini' % (modName, suffix)), 'w') as conf:
+        cfg.write(conf)
+
+
+
+def readCfgFile(avaDir, module, suffix=''):
```
fso42, line 182: This function is not really needed as far as I can see, as the function: `cfgUtils.getModuleConfig` 
allows for a fileOverride in which a specified file can be used.

avaframe/in3Utils/cfgUtils.py
```diff
@@ -11,6 +11,7 @@
 # Local imports
 import avaframe as avaf
 from avaframe.in3Utils import logUtils
+import pickle
```
fso42, line 14: Not needed?

avaframe/out3Plot/plotUtilsCfg.ini
```diff
@@ -45,5 +45,5 @@ unitrelTh = m
 # name for result parameters
 nameppr = peak pressure
 namepfd = peak flow depth
-namepv = peak speed
+namepv = peak flow velocity
```
fso42, line 48:
```suggestion
namepv = peak velocity
```

fso42 (changes requested):

avaframe/in3Utils/cfgUtils.py
awirb: I would suggest to keep it as a separate function as the writeCfgFile creates one that is called moduleName_settings.ini, and the read by default reads the moduleName_settings.ini so it appends the correct ending and I think in this way it is less likely to mix it with reading the standard/default moduleNameCfg.ini file..

avaframe/in3Utils/cfgUtils.py
awirb: True, thanks

avaframe/in1Data/computeFromDistributionCfg.ini
awirb: Done

avaframe/out3Plot/plotUtilsCfg.ini
awirb: keep this for now

fso42 (approved):

PR #259, opened 2021-01-04
fix #255 call explicitly pU

### LINKED ISSUE #255: use explicit import for plotUtils functions and parameters
remove from avaframe.out3Plot.plotUtils import * and use explicit imports

matthiasto:
from now on, import plotUtils as following:
import avaframe.out3Plot.plotUtils as pU
and use:
pU.functionName() or pU.variableName to cal the functions and variables defined in plotUtils.
I changed it everywhere I think but did not check if it works properly in probAna.


awirb:
it works fine for probAna

fso42 (approved):

PR #265, opened 2021-01-05
change ent and res flag logic

### LINKED ISSUE #264: Handling of ini-flag ent and res
ATM if `ent=True` and no entrainment area is given, only a warning is given. 

This might lead to the confusion of setting `ent=True` and expecting entrainment to happen (without providing an entrainment shapefile; res ditto)

So I suggest renaming these flags to something like `turnOffEnt` or `forceNoEnt` for developing purposes. 

The logic from an end users perspective should be:
- If an entrainment shape file is provided with the input, then entrainment is used, otherwise no entrainment takes place

Dev/Powerusers however can still quickly turn off ent/res by using the flags

awirb:
fix #264

### DIFF
--- a/avaframe/com1DFA/com1DFA.py
+++ b/avaframe/com1DFA/com1DFA.py
@@ -126,7 +126,7 @@ def com1DFAMain(cfg, avaDir):
     if flagNoEnt:
         log.info('Entrainment is forced off')
     if flagNoRes:
-        log.info('Resistance is turned off')
+        log.info('Resistance is forced off')
 
     # Log current avalanche directory
     log.debug('Your current avalanche name: %s' % avaDir)
--- a/avaframe/com1DFA/com1DFA.py
+++ b/avaframe/com1DFA/com1DFA.py
@@ -112,8 +112,8 @@ def com1DFAMain(cfg, avaDir):
     cfgGen = cfg['GENERAL']
     com1Exe = cfgGen['com1Exe']
     modName = 'com1DFA'
-    flagEnt = cfgGen.getboolean('flagEnt')
-    flagRes = cfgGen.getboolean('flagRes')
+    flagNoEnt = cfgGen.getboolean('flagNoEnt')
+    flagNoRes = cfgGen.getboolean('flagNoRes')
     fullOut = cfgGen.getboolean('flagOut')
     cfgPar = cfg['PARAMETERVAR']
     resDir = os.path.join(avaDir, 'Work', 'com1DFA')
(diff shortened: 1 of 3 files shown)


avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -8,10 +8,10 @@
...
 # True if full output shall be printed to terminal
 flagOut = False
-# True if entrainment areas shall be included
-flagEnt = True
-# True if resistance areas shall be included
-flagRes = True
+# True if entrainment areas shall NOT be included
+flagNoEnt = True
+# True if resistance areas shall NOT be included
+flagNoRes = True
```
fso42, line 14: I think they should be `False` by default:
This leads to the expected behaviour of including entrainment if an entrainment shapefile is present, and not including it if none is present...

```suggestion
# True if entrainment areas shall NOT be included
flagNoEnt = False
# True if resistance areas shall NOT be included
flagNoRes = False
```

avaframe/com1DFA/com1DFA.py
```diff
@@ -123,7 +123,10 @@ def com1DFAMain(cfg, avaDir):
     defValues = cfg['DEFVALUES']
 
     # Log chosen settings
-    log.debug('The chosen settings: entrainment - %s , resistance - %s ' % (flagEnt, flagRes))
+    if flagNoEnt:
+        log.info('Entrainment is turned off')
```
fso42, line 127:
```suggestion
        log.info('Entrainment is forced off')
```

avaframe/com1DFA/com1DFA.py
```diff
@@ -123,7 +123,10 @@ def com1DFAMain(cfg, avaDir):
...
+    if flagNoRes:
+        log.info('Resistance is turned off')
```
fso42, line 129:
```suggestion
        log.info('Resistance is forced off')
```

avaframe/in1Data/getInput.py
```diff
@@ -77,10 +77,13 @@ def getInputData(avaDir, cfg, flagDev=False):
...
     # Initialise resistance areas
-    if cfg.getboolean('flagRes'):
+    if cfg.getboolean('flagNoRes'):
+        resFiles = []
+        resFiles.append('')
+    else:
         resFiles = glob.glob(inputDir+os.sep + 'RES' + os.sep+'*.shp')
         if len(resFiles) < 1:
-            log.warning('No resistance file')
+            log.debug('No resistance file')
```
fso42, line 86:
```suggestion
            log.debug('No resistance file found')
```

avaframe/in1Data/getInput.py
```diff
@@ -89,15 +92,15 @@ def getInputData(avaDir, cfg, flagDev=False):
...
     # Initialise entrainment areas
-    if cfg.getboolean('flagEnt'):
+    if cfg.getboolean('flagNoEnt'):
+        entFiles = []
+        entFiles.append('')
+    else:
         entFiles = glob.glob(inputDir+os.sep + 'ENT' + os.sep+'*.shp')
         if len(entFiles) < 1:
-            log.warning('No entrainment file')
+            log.debug('No entrainment file')
```
fso42, line 103:
```suggestion
            log.debug('No entrainment file found')
```

fso42 (changes requested):

fso42 (approved):

PR #268, opened 2021-01-07
Add option to have underscore in release area naming

### LINKED ISSUE #262: Handle underscores in input shapefiles dems etc
How should we handle underscores in:
- release area files?
- DEMs?
- entrainment files?

Treat all the same? 
Remove the underscores?
Throw an error and demand renaming?
Rename automatically?

Problem is the seperation in a multivalue ini parameter -> right now it is seperated by an underscore

awirb:
if _ appears in release area name - _AF is added for the file and log naming in com1DFA
* attempt to fix #262 
* adjust ini files in benchmarks


avaframe/com1DFA/com1DFA.py
```diff
@@ -164,6 +165,11 @@ def com1DFAMain(cfg, avaDir):
         # Set release areas and simulation name
         relName = os.path.splitext(os.path.basename(rel))[0]
         simName = relName
+        if '_' in relName:
+            flagBadName = True
+            log.warning('Release area scenario file name includes an underscore - bad naming \
+            the suffix _AF will be added')
```
fso42, line 171:
```suggestion
            log.warning('Release area scenario file name includes an underscore \
            the suffix _AF will be added')
```

avaframe/in1Data/getInput.py
```diff
@@ -60,7 +60,8 @@ def getInputData(avaDir, cfg, flagDev=False):
     elif cfg['releaseScenario'] != '':
         releaseDir = 'REL'
         relFiles = []
-        releaseFiles = cfg['releaseScenario'].split('_')
+        releaseFiles = cfg['releaseScenario'].split('|')
```
fso42, line 63: Please use the splitIniValueToArray function here...

avaframe/in3Utils/fileHandlerUtils.py
```diff
@@ -351,13 +351,23 @@ def makeSimDict(inputDir, varPar='', avaDir=''):
...
+            data['resType'].append(nameParts[3])
+            data['simName'].append(nameParts1[0] + '_AF_' + ('_'.join(nameParts[0:3])))
+        else:
+            nameParts = name.split('_')
+            data['releaseArea'].append(nameParts[0])
+            data['simType'].append(nameParts[1])
+            data['modelType'].append(nameParts[2])
+            data[varPar].append(nameParts[3])
+            data['resType'].append(nameParts[4])
+            data['simName'].append('_'.join(nameParts[0:4]))
```
fso42, line 370:
```suggestion
        if '_AF_' in name:
            nameParts = name.split('_AF_')
            fNamePart = nameParts[0] + '_AF'
            infoParts = nameParts[1].split('_')

        else:
            nameParts = name.split('_')
            fNamePart = nameParts[0]
            infoParts = nameParts[1:]

        data['releaseArea'].append(fNamePart)
        data['simType'].append(infoParts[0])
        data['modelType'].append(infoParts[1])
        data[varPar].append(infoParts[2])
        data['resType'].append(infoParts[3])
        data['simName'].append(fNamePart + '_' + ('_'.join(infoParts[0:3])))

```
Please test, it gives the same results for me...

fso42 (changes requested):

matthiasto (approved):

avaframe/in1Data/getInput.py
awirb: the splitIniValueToArray is not supposed to work for this case, as a string is required for the file name not a numpy array which is returned in splitIniValueToArray, I could make an additional function but so far this is the only case where we need this and it is one line so I'd suggest to keep it here

fso42 (approved):

PR #276, opened 2021-01-19
Pretify dfa py

matthiasto:


fso42:
Computation of runCom1DFA.py fails if
`contCMap` in `plotUtilsCfg.ini  = False`
which is the default. 
Please make sure to not rely on the config if your data producing routine does not provide discrete data! -> set continous to True in com1DFA.py line 1487....

docs/2_5DSPH.rst
```diff
@@ -0,0 +1,187 @@
...
+
+.. math::
+   W_{ij} = W(\mathbf{x_i},\mathbf{x_j},r_0) = W(\mathbf{x_i}-\mathbf{x_j},r_0) = W(\mathbf{r_{ij}},r_0) = \frac{10}{\pi r_0^5}\left\{
+   \begin{aligned}
+   & (r_0 - \left\Vert \mathbf{r_{ij}}\right\Vert)^3, \quad &0\leq \left\Vert \mathbf{r_{lj}}\right\Vert \leq  r_0\\
+   & 0 , & r_0 <\left\Vert \mathbf{r_{ij}}\right\Vert
+   \end{aligned}
+   \right.
+   :label: kernel function
+
```
fso42, line 69: Please split this equation, it is getting too wide on the docs. 
image

Same goes for equations 9, 13 of the com1DFA Kernel theory

fso42 (changes requested):
Following issues:

- Please see my comment outside of this review
- You adjusted the GeneralPlotHeader.tex, but did not provide the tex file to produce the new plot? (Fig 8 in docs)

awirb (changes requested):

PR #285, opened 2021-01-27
Fix bug and update input dir

awirb:
bug - export last time step of com1DFAPy results by calling it from the configuration file was not working properly


fso42 (changes requested):
Please add a description of what '_the bug_' was to make it searchable/understandable (either in the title or description)

fso42 (approved):

PR #289, opened 2021-02-01
Config2 dict

matthiasto:
Switching to dictionary for passing config arguments in the con1DFA


matthiasto:
We choose to stick to config parser

Issue #298, opened 2021-02-09
Remove showplot Flag from generate Topo cfgs - Use main showPlot cfg


fso42:
could be changed to use the flag in the general configuration file and remove the flag in the generateTopoCfg.ini

_Originally posted by @awirb in https://github.com/avaframe/AvaFrame/pull/295#discussion_r571873409_

Change in in3Utils/generateTopoCfg.ini and in3Utils/getReleaseAreaCfg.ini and adjust plotGeneratedDEM 

While at it, also make  plotGeneratedDEM  work with only main cfg maybe (not the DEM cfg)

awirb:
fixed by #305

PR #305, opened 2021-02-10
Add docstrings

awirb:
solve #298 use global flag to show plots in plotgeneratedDEM


fso42 (changes requested):
I think all of the showplot (small `p`) occurrences should be removed in the code -> cfg files and runscripts...

PR #313, opened 2021-02-16
adding first pytests for DFA

matthiasto:
Added new tests for the cython code
I changed the workflow, not sure it was the best way to do it


fso42 (changes requested):
As discussed please address the pytest flag to separate testing and debugging

fso42 (approved):

PR #317, opened 2021-02-17
Compare dfa

matthiasto:


fso42:
Since we are now touching this again, maybe we should rethink whether it is a good idea to use our own functions for this, instead of eg. something like:

https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RectBivariateSpline.html#scipy.interpolate.RectBivariateSpline

with scipy already being in our requirements...
This means less code for us and it's easier to test different interpolation methods....

fso42:
As an example (quick and dirty) consider following code. Those alternatives would mean we could drop a substantial amount of code in geoTrans **and** be more flexible / have access to more interpolations ....

```
import numpy as np
import time
# Local imports
import avaframe.in3Utils.geoTrans as geoTrans
import avaframe.in2Trans.ascUtils as IOf
from scipy.interpolate import RectBivariateSpline
from scipy.interpolate import interp2d
from scipy.interpolate import RegularGridInterpolator
import matplotlib.pyplot as plt

dem = {}
Points = {}
header = IOf.cASCheader()
header.xllcenter = 0
header.yllcenter = 0
header.cellsize = 1

rasterdata = np.array([[0, 1, 2, 3], [1, 2, 3, 4], [2, 3, 4, 5], [3, 4, 5, 6]])
x = np.array([0, 1, 2, 3])
y = np.array([0, 1, 2, 3])

# Handling of out of bounds is not tested yet !!
Points['x'] = np.array([0.4 , 0 , 0.5, 1.6, 2.4, 2.5])
Points['y'] = np.array([0 , 1.4, 0.5, 0.6, 0.4, 1])

dem['header'] = header
dem['rasterData'] = rasterdata

# ===========CURRENT
Points, _ = geoTrans.projectOnRasterVect(dem, Points, interp='bilinear')
print(Points['z'])

# more info find at: https://docs.scipy.org/doc/scipy/reference/interpolate.html#module-scipy.interpolate

#  ==========THIS (no nearest option? )
RBS = RectBivariateSpline(x, y, np.transpose(rasterdata), ky=1, kx=1)
print(RBS(Points['x'],Points['y'],grid=False))

# ===========Or This (not pointwise yet)
I2D = interp2d(x, y, np.transpose(rasterdata), kind='linear', fill_value = np.NaN)
print(I2D(Points['x'][0] ,Points['y'][0]))

#============== or this
RGI = RegularGridInterpolator((x, y), rasterdata)
print(RGI((Points['x'] ,Points['y']), method = 'linear'))

# print(rasterdata.shape)
# X, Y = np.meshgrid(x, y)
# # rasterdata = np.transpose(rasterdata)
# plt.pcolormesh(X, Y, rasterdata, shading='auto')
# plt.plot(Points['x'], Points['y'], "ok", label="input point")
# plt.colorbar()
# plt.show()

```

fso42:
Regarding my last comment: 
we decided to:
- keep our own interpolation routines for the frequent calls from eg. com1DFApy
- streamline geoTrans.py (eg only one pointToRaster that does not care whether a vector or single point is provided)
- use scipy interpolation where more flexibility/functionality is needed

PR #336, opened 2021-02-26
Add description file logic to benchmarks

### LINKED ISSUE #332: Restructure Benchmarks
* one directory for each benchmark 
* add benchmark description files

awirb:
* rename benchmark test folders 
-> so now one folder for each benchmark test 
TODO
some tests still need to be renamed!
* now filtering of tests using TAGS, Types,...
* fix #332


matthiasto:
When you run the standard tests, the wogNull for example, both entres and null are run. Could we change the ini so that only the null is run?

fso42 (approved):

PR #372, opened 2021-03-25
update FlatPlane test

matthiasto:


avaframe/runTestFP.py
```diff
@@ -31,7 +31,7 @@
 modName = 'com1DFAPy'
 
 # Clean input directory(ies) of old work and output files
-initProj.cleanSingleAvaDir(avalancheDir, keep=logName)
+# initProj.cleanSingleAvaDir(avalancheDir, keep=logName)
```
fso42, line 34: Is this comment intentional?

matthiasto: yep otherwise you delete all the data in the other com modules. there is already a clean module called after in the runCom1DFAPy

PR #374, opened 2021-03-26
Improving aimec plots

matthiasto:
adding contour line comparison plot and additional information to the mass balance plot


awirb (changes requested):
Suggestion: add start of runout area as legend to the AreaComparisonToReference plots

PR #380, opened 2021-04-06
Set string to float

awirb:


avaframe/com1DFA/com1DFA.py
```diff
@@ -182,7 +182,10 @@ def com1DFAMain(cfg, avaDir):
         relDict = sP.SHP2Array(rel)
         for k in range(len(relDict['d0'])):
             if relDict['d0'][k] == 'None':
-                relDict['d0'][k] = '1.0'
+                relDict['d0'][k] = defValues.getfloat('RelTh')
+            else:
+                relDict['d0'][k] = float(relDict['d0'][k])
```
matthiasto, line 187: If there is already a value in here, why is it not already a float?

PR #401, opened 2021-04-19
Cfg com1dfa py

matthiasto:
Add some options to the com1DFAPy cfg to be able to enable disable sph and curvature
Change th way to compute the l and s distance (only used for testing the energy analysis stuff)

### DIFF
--- a/avaframe/com1DFAPy/DFAfunctionsCython.pyx
+++ b/avaframe/com1DFAPy/DFAfunctionsCython.pyx
@@ -479,7 +479,7 @@ def updatePositionC(cfg, particles, dem, force):
   cdef double[:, :] Nx = dem['Nx']
   cdef double[:, :] Ny = dem['Ny']
   cdef double[:, :] Nz = dem['Nz']
-  cdef double[:, :] Z = dem['rasterData']
+  cdef double[:, :] ZDEM = dem['rasterData']
   # initializeinter = np.zeros(N, dtype=np.float64)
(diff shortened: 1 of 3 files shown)


avaframe/com1DFAPy/com1DFACfg.ini
```diff
@@ -13,9 +13,14 @@ rhoEnt = 100
 massPerPart = 1250
 # minimum flow depth [m]
 hmin = 0.05
+# curvature acceleration coefficient
+# take curvature term into account in the gravity acceleration term
+# 0 if deactivated, 1 if activated
+curvAcceleration = 1
```
awirb, line 20: Would it be an option to put things like this in a Section 'Development'?

avaframe/com1DFAPy/DFAfunctionsCython.pyx
```diff
@@ -259,7 +260,7 @@ def computeForceC(cfg, particles, fields, dem, dT):
       accNormCurv = (ux*(nxEnd-nx) + uy*(nyEnd-ny) + uz*(nzEnd-nz)) / dt
       # normal component of the acceleration of gravity
       gravAccNorm = - gravAcc * nzAvg
-      effAccNorm = gravAccNorm + accNormCurv
+      effAccNorm = gravAccNorm + curvAcceleration * accNormCurv
```
awirb, line 263: maybe add a comment here that this is for development purposes? or that it is to turn off curvature?

awirb (changes requested):

PR #406, opened 2021-04-21
Remesh inputs

matthiasto:
remesh input raster using rectBivariantSplines
allow cfl time stepping

### DIFF
--- a/avaframe/in3Utils/geoTrans.py
+++ b/avaframe/in3Utils/geoTrans.py
@@ -4,10 +4,14 @@
 import logging
 import math
 import numpy as np
+import scipy as sp
 import copy
+from matplotlib import pyplot as plt
 
 # Local imports
 import avaframe.in2Trans.ascUtils as IOf
+import avaframe.out3Plot.makePalette as makePalette
+import avaframe.out3Plot.plotUtils as pU
 
 # create local logger
 log = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


avaframe/in3Utils/geoTrans.py
```diff
@@ -149,6 +150,64 @@ def resizeData(raster, rasterRef):
...
+    xllcenter = headerDEM.xllcenter
+    yllcenter = headerDEM.yllcenter
+    cszDEM = headerDEM.cellsize
+    # remesh if input DEM size does not correspond to the computational cellSize
+    if np.abs(cszNew - cszDEM) > cszThreshold:
+        log.info('Remeshing the input DEM (of cell size %.2g m) to a cell size of %.2g m' % (cszDEM, cszNew))
+        x = np.linspace(0, (nColsDEM-1)*cszDEM, nColsDEM) + xllcenter
+        y = np.linspace(0, (nRowsDEM-1)*cszDEM, nRowsDEM) + yllcenter
+        z = dem['rasterData']
+        fInterp = sp.interpolate.RectBivariateSpline(x, y, np.transpose(z), ky=3, kx=3)
```
fso42, line 187: I absolutely like the simplicity of using RectBivariateSpline, but I'm undecided wheter griddata would give more options (like bilinear, cubic etc interpolation). Question is: do we want those options? Or should we just put a comment beforehand that this would be the spot to add griddata if the need ever arises?

fso42 (approved):

avaframe/in3Utils/geoTrans.py
matthiasto: you can here change the degree of the spline which ends up in kind of the same result I think

fso42: Ok, then: mention in the docstring that RectBivar... is used, and that this would be the place to swap to another function like griddata or similar if the need arises

PR #417, opened 2021-04-27
make remesh work with noData

matthiasto:
Here is a fix to the remeshing function for the case of dem with no data areas.
I will try to add some pytest to check it but in the mean time you can use this for the standard tests.

### DIFF
--- a/avaframe/in3Utils/geoTrans.py
+++ b/avaframe/in3Utils/geoTrans.py
@@ -152,10 +152,10 @@ def resizeData(raster, rasterRef):
 
 def remeshDEM(cfg, dem):
     """ change DEM cell size by reprojecting on a new grid
-    
-    Interpolation is based in RectBivariateSpline of order 3. Here would be the place
+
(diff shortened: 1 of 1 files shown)


avaframe/in3Utils/geoTrans.py
```diff
@@ -204,9 +212,10 @@ def remeshDEM(cfg, dem):
         dem['header'] = headerRemeshed
         xNew = np.linspace(0, (nColsRemeshed-1)*cszNew, int(nColsRemeshed)) + xllcenter
         yNew = np.linspace(0, (nRowsRemeshed-1)*cszNew, int(nRowsRemeshed)) + yllcenter
+        xNewGrid, yNewGrid = np.meshgrid(xNew, yNew)
+        zNew = sp.interpolate.griddata((xGrid, yGrid), z, (xNewGrid, yNewGrid), method='cubic', fill_value=headerDEM.noDataValue)
```
fso42, line 216: Should we expose the `method` as configuration parameter?

fso42 (approved):

PR #421, opened 2021-04-27
Stand tests update

awirb:
* update of benchmarks regarding the new simTypes, works for com1DFAPy for now
* TODO adapt com1DFA simulation naming - Done with newest commit 
* also updated variations test 
* now com1DFA also uses simTypes ent, res, entres, null and available


fso42 (approved):

avaframe/com1DFA/com1DFA.py
```diff
@@ -93,6 +93,71 @@ def copyReplace(origFile, workFile, searchString, replString):
         file.write(fileData)
 
 
+def getSimulation(cfg, rel, entResInfo):
+    """ get Simulation to run for a given release
```
fso42, line 97: Bit unclear what is meant here. Maybe 
Given a simulation scenario, returns the types of simulations that need to be performed

avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -8,10 +8,12 @@
...
 # True if full output shall be printed to terminal
 fullOut = False
-# True if entrainment areas shall NOT be included
-noEntrainment = False
-# True if resistance areas shall NOT be included
-noResistance = False
+
+#++++++++++++++++ simulation type
+# list of simulations that shall be performed (null, ent, res, entres, available (use all available input data))
+simTypeList = null|available
```
fso42, line 14: should be only `available`, since `available` contains `null`...

fso42 (changes requested):

avaframe/com1DFA/com1DFA.py
awirb: I changed this to: Get a list of all simulations that shall be performed according to simTypeList in configuration file;
        and a dictionary with information on release area

avaframe/com1DFA/com1DFACfg.ini
awirb: correct, set it to available

fso42 (approved):

PR #423, opened 2021-04-28
Fix comparestandalone py

matthiasto:


avaframe/runComparisonModules.py
```diff
@@ -92,21 +92,32 @@
...
             simType = 'entres'
-        com1DFASimName = reportDcom1DFA['simName']['name']
+            reportDcom1DFA = reportDcom1DFAentres
+        elif reportDcom1DFAent:
+            simType = 'ent'
+            reportDcom1DFA = reportDcom1DFAent
+        elif reportDcom1DFAres:
+            simType = 'res'
+            reportDcom1DFA = reportDcom1DFAres
+
```
awirb, line 118: why not set the simType in the first if ajd reportDcom1DFA directly there too?

awirb (changes requested):

PR #437, opened 2021-05-18
fix #389 to be able to use custom parameters in AB

### LINKED ISSUE #389: AB custom parameter set
**Describe the bug**
Not possible to use a custom parameter set

**Additional context**
All possible ini parameters need to be define in the ini file (not only in the local ini file)
We forgot to define the customParam dictionary

matthiasto:


avaframe/com2AB/com2ABCfg.ini
```diff
@@ -8,16 +8,14 @@
 [ABSETUP]
 # if small avalanche set is wanted
 smallAva = False
- # if custom avalanche set is not wanted
+ # if custom avalanche set is not wanted switch to True and specify your parameters
```
fso42, line 11:
```suggestion
 # if custom avalanche set is wanted switch to True and specify your parameters
```

fso42 (changes requested):

fso42 (approved):

PR #445, opened 2021-05-26
adding mass based stop criterion

matthiasto:
added stop criterion based on flowing mass:
stops simulation when the flowing mass (all particles going faster than 0.1 m/s) is smaller then the peakFlowingMass


avaframe/com1DFAPy/DFAfunctionsCython.pyx
```diff
@@ -586,6 +586,9 @@ def updatePositionC(cfg, particles, dem, force):
       uyNew = uyNew * uMag / (uMagNew + velMagMin)
       uzNew = uzNew * uMag / (uMagNew + velMagMin)
 
+    if uMag > 0.01:
```
fso42, line 589: This setting needs to go to the ini file

fso42 (changes requested):

avaframe/com1DFAPy/DFAfunctionsCython.pyx
matthiasto: agreed, how should I call it? I just did not like the fact that for the mass stopping criterion you need an extra input parameter.

fso42 (approved):

PR #459, opened 2021-06-09
allow adding section to cfg from override ini

awirb:
* if in local_ or overrid_cfg file an additional section, when compared to the default cfg, exists, this entire section is added, however if an additional key exists in a section that also exists in the default cfg - this key is ignored
* update damBreak and simiSol ini files

### DIFF
--- a/avaframe/runDamBreak.py
+++ b/avaframe/runDamBreak.py
@@ -5,6 +5,7 @@
 # Load modules
 import os
 import numpy as np
+import pathlib
 
 # Local imports
 import avaframe.com1DFA.com1DFA as com1DFA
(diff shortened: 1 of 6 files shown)


matthiasto (approved):

avaframe/runDamBreak.py
```diff
@@ -47,7 +47,7 @@
 Particles, Fields, Tsave, dem, plotDict, reportDictList = runCom1DFA.runCom1DFA(avaDir=avaDir, cfgFile=damBreakCfg)
 
 # create simDict of results
-inputDir = 'data/avaDamBreak/Outputs/com1DFAPy/peakFiles/timeSteps'
+inputDir = 'data/avaDamBreak/Outputs/com1DFA/peakFiles/timeSteps'
```
fso42, line 50: This fails on windows. I suggest we start using pathlib for stuff like this (as long as it is in a run script)

fso42 (changes requested):

avaframe/runDamBreak.py
awirb: switched to pathlib

PR #463, opened 2021-06-11
fixing find beta angle issue 

matthiasto:
Finding the beta angle did not work when the profile is very finely resampled or has uneven steps.
We now use a given threshold defined in the ini file to determine how many point (how much distance) has to be bellow the Angle in order to define a point as beta point

### DIFF
--- a/avaframe/in3Utils/geoTrans.py
+++ b/avaframe/in3Utils/geoTrans.py
@@ -427,11 +427,6 @@ def isCounterClockWise(path):
       path: matplotlib.path
         polygon path
     Returns
-    -------Parameters
-    ----------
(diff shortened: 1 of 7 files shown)


avaframe/in3Utils/geoTrans.py
```diff
@@ -394,6 +427,11 @@ def isCounterClockWise(path):
       path: matplotlib.path
         polygon path
     Returns
+    -------Parameters
+    ----------
```
fso42, line 431: formatting error?

fso42 (approved):

awirb (changes requested):

PR #470, opened 2021-06-17
Prepare runOperational for QGis connector

fso42:
- make runOperational callable
- Function to write com2AB files to shapefiles
- change some log from info to debug in com1DFA (see #177)


avaframe/runOperational.py
```diff
@@ -1,71 +1,102 @@
...
+
+    Parameters
+    ----------
+    avalancheDir: str
+        path to avalanche directory (setup eg. with init scipts)
+
+    Returns
+    -------
+    abShpFile: str
+        path to com1AB results as shapefile
```
awirb, line 33: com2AB

avaframe/runOperational.py
```diff
@@ -1,71 +1,102 @@
...
+    peakFilesDF: pandas dataframe
```
awirb, line 34: suggestion add that this is for com1DFA results

awirb (changes requested):

PR #486, opened 2021-07-06
add pytest for configurationfiles utils

awirb:
* add check if folder exists, if input parameters are in correct format
*  add check if configuration files are found

### DIFF
--- a/avaframe/tests/test_cfgUtils.py
+++ b/avaframe/tests/test_cfgUtils.py
@@ -84,6 +84,21 @@ def test_convertDictToConfigParser():
     assert cfgDict['TEST1']['mu'] == cfg['TEST1'].getfloat('mu')
 
 
+def test_orderSimFiles():
+    """ test generating order of simulation results """
+
+    avaTestDir = 'avaHockeyChannelPytest'
+    dirPath = os.path.dirname(__file__)
+    avaDir = os.path.join(dirPath, '..', '..', 'benchmarks', avaTestDir)
+    inputDir = os.path.join(avaDir, 'Outputs', 'com1DFA', 'peakFiles')
+
+    varParList = 'releaseScenario'
+
+    simFilesDF = cfgUtils.orderSimFiles(avaDir, inputDir, varParList, True, specDir='')
+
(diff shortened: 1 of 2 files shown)


avaframe/tests/test_cfgUtils.py
```diff
@@ -52,3 +55,83 @@ def test_cfgHash():
...
+    simNames = cfgUtils.filterSims(avaDir, parametersDict, specDir='')
+
+    testRel = False
+    testRel2 = False
+    if 'release1HS' in simNames[0]:
+        testRel = True
+    if 'release2HS' in simNames[0]:
+        testRel2 = True
+
+    assert testRel == True
```
matthiasto, line 135: you dont need the == True for a boolean, assert testRel should be enough

matthiasto (approved):

avaframe/tests/test_cfgUtils.py
```diff
@@ -52,3 +55,83 @@ def test_cfgHash():
...
+    assert cfgDict['TEST1']['mu'] == cfg['TEST1'].getfloat('mu')
+
+
+def test_orderSimFiles():
+    """ test generating order of simulation results """
+
+    avaTestDir = 'avaHockeyChannelPytest'
+    dirPath = os.path.dirname(__file__)
+    avaDir = os.path.join(dirPath, '..', '..', 'benchmarks', avaTestDir)
+    inputDir = os.path.join(avaDir, 'Outputs', 'com1DFA', 'peakFiles')
```
fso42, line 93: I suggest switching all the os.path.joins to pathlib functions (also further down)

fso42 (changes requested):

avaframe/tests/test_cfgUtils.py
awirb: now all paths in here using pathlib

avaframe/tests/test_cfgUtils.py
awirb: done

fso42 (approved):

PR #500, opened 2021-07-27
Add pytests and update getStats

awirb:
* add pytests for fileHandlerUtils and getStats
* add option to restrict the computation of statistical measures to a specific result field - for example only compute max values of peak flow velocity where peak flow depth > 0 
* add computation of mean, min and std values of result fields 
* update statsPlots x and y label to indicate if max, mean, min values are plotted 
* update plots in docu


avaframe/ana4Stats/getStats.py
```diff
@@ -82,35 +54,53 @@ def extractMaxValues(inputDir, cfgMain, avaDir, varPar, nameScenario='', paramet
...
 
             # Load data
             fileName = peakFilesDF['files'][m]
             simName = peakFilesDF['simName'][m]
-            data = np.loadtxt(fileName, skiprows=6)
-
-            # compute max
-            max = np.amax(data)
-
+            dataFull = IOf.readRaster(fileName)
```
matthiasto, line 73: read dataFull only in the else to avoid overwriting it in the if

avaframe/ana4Stats/getStats.py
```diff
@@ -82,35 +54,53 @@ def extractMaxValues(inputDir, cfgMain, avaDir, varPar, nameScenario='', paramet
...
+
+            # if restrictType, set result field values to nan if restrictType result field equals 0
+            if restrictType != '':
+                peakFilesSimName = peakFilesDF[peakFilesDF['simName'] == simName]
+                fileNamePFD = peakFilesSimName[peakFilesSimName['resType'] == restrictType]['files'].values[0]
```
matthiasto, line 78: what does PFD stand for?

avaframe/tests/test_getStats.py
```diff
@@ -0,0 +1,95 @@
...
+            for m in range(M):
+                line = np.array([resultArray[m, :]])
+                np.savetxt(outFile, line, fmt='%.16g')
+
+            outFile.close()
+
+        cfg = configurationSettings[k]
+        cfg.optionxform = str
+        with open(os.path.join(configFiles[k]), 'w') as conf:
+            cfg.write(conf)
```
matthiasto, line 81: I am not sure it is a good Idea to do this here. Would it for example be possible to create the files first on your computer and add them to the data folder in tests? or is it a bad idea

matthiasto (changes requested):

avaframe/ana4Stats/getStats.py
awirb: dataFull is required in both cases, it is the actual result field, inside the if the dataFull is restricted to where dataPFD > 0, inside the else it is not restricted

avaframe/ana4Stats/getStats.py
awirb: peak flow depth, I called it like this as if pfd is zero - there was no ava so this field is useful to restrict the other result fields too

avaframe/tests/test_getStats.py
awirb: Is it better in terms of computational costs, or what is the reason not to do this here?

matthiasto: just makes the pytest cleaner and easier to read I would say

avaframe/ana4Stats/getStats.py
matthiasto: true I did not see it...

avaframe/ana4Stats/getStats.py
matthiasto: but where do I see that you reed the PFD?

awirb: I will rename it, it is not PFD you can choose what it is in the ini file restrictType=pfd for example.. by default it is pfd

matthiasto: okay now I understand

PR #504, opened 2021-07-29
Change dtSave array

### LINKED ISSUE #287: getTimeIndex doesnt work properly if dtSave < actual dt [com1DFAdev]
if dtSave < than actual time step getTimeIndex doesn't work properly as the indices do not correspond to Tsave in the way it is assumed by getTimeIndex for now

function getTimeIndex in in3Utils/fileHandlerUtils

awirb:
now a saving time step array is created and is updated according to actual time step - all values smaller than actual time step are removed - 
fix #287 

* change in placeParticles defining number of particles per cell for uniform and random 
* add small pytests for com1DFA

### DIFF
--- a/avaframe/tests/test_com1DFA.py
+++ b/avaframe/tests/test_com1DFA.py
@@ -80,29 +80,29 @@ def test_initializeMesh():
     demHeader = IOf.cASCheader()
     demHeader.xllcenter = 101.23
     demHeader.yllcenter = 24.54
-    demHeader.cellsize = 5.0
+    demHeader.cellsize = 1.0
     demHeader.noDataValue = -9999
     demHeader.nrows = 5
     demHeader.ncols = 5
 
     # define plane with constant slope of 45°
-    demData = np.asarray([[1., 2., 3., 4., 5.],
-                          [1., 2., 3., 4., 5.],
-                          [1., 2., 3., 4., 5.],
-                          [1., 2., 3., 4., 5.],
-                          [1., 2., 3., 4., 5.]])
+    demData = np.asarray([[1., 2., 3., 4., np.nan],
+                          [1., 2., 3., 4., np.nan],
+                          [1., 2., 3., 4., np.nan],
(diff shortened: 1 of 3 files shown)


avaframe/tests/test_com1DFA.py
```diff
@@ -1,112 +1,27 @@
...
-    # Run Standalone DFA
-    reportDictList = com1DFA.com1DFAMain(cfg, avaDir)
-    reportD = reportDictList[0]
-    pprCom1DFA = np.loadtxt(os.path.join(avaDir, 'Outputs', 'com1DFA', 'peakFiles',
-                                         reportD['simName']['name'] + '_ppr.asc'), skiprows=6)
+    # setup required input
+    class testHeader:
+        def __init__(self):
+            self.xllcenter = 1.0
+            self.yllcenter = 4.0
```
matthiasto, line 19: why not simply define a new header and set the x and y?

matthiasto (approved):

avaframe/tests/test_com1DFA.py
```diff
@@ -1,112 +1,211 @@
...
+    maskTest[0, 0:2] = 1.2
+    maskTest[1, 0] = 1.2
+
+    # call function to be tested
+    Mask2 = com1DFA.polygon2Raster(demHeader, Line, 0.1, th=th)
+    maskTest2 = maskTest.copy()
+    maskTest2[1, 1] = 1.2
+
+    assert np.array_equal(maskTest, Mask)
+    assert np.array_equal(maskTest2, Mask2)
```
matthiasto, line 194: maybe add a case where there are two polygons that overlap and check that we get the good warning or error...

matthiasto (changes requested):

avaframe/tests/test_com1DFA.py
awirb: this is for prepareArea then, next PR

PR #505, opened 2021-07-30
inerpOption from ini for initialization of Hpart

matthiasto:


awirb (approved):
where else do we need the function from geoTrans for this?

PR #511, opened 2021-08-02
move debug plots ot of com1DFA

matthiasto:
There are no calls to matplotlib for plotting purposes in com1DFA
I hope this removes the need to call for matplotlib.use('agg') thing when using qgis


awirb (approved):

snowtechblog:
I came across the nuisance with the matplotlib backend 'agg' because some functions import the outDebugPlots.py even though no debugPlots are requested.

Perhaps one could wrap their import statement into a if-clause
```
if debugPlot:
    import avaframe.out3Plot.outDebugPlots as debPlot
```
 in `com1DFA.py`, line 47, and in `particleInitialisation.py`, line 23.

PR #514, opened 2021-08-03
add pytest for out1PeakPlot

awirb:
* check also for pathlib path in out3Plot putAvaNameOnPlot 
* remove unused inputs


matthiasto (commented):
Can you not use a .asc file in benchmark that already exist?

PR #518, opened 2021-08-04
Test dfa tools

matthiasto:


avaframe/com1DFA/DFAtools.py
```diff
@@ -510,30 +469,25 @@ def normalize(x, y, z):
...
             x component of the normalized vector
-        yn: numpy array
+        y: numpy array
             y component of the normalized vector
-        zn: numpy array
+        z: numpy array
             z component of the normalized vector
     """
     # TODO : avoid error message when input vector is zero and make sure
     # to return zero
```
awirb, line 480: so this is solved no?

avaframe/tests/test_DFAtools.py
```diff
@@ -0,0 +1,232 @@
+"""Tests for module com1DFAtools"""
```
awirb, line 1: maybe put correct name here

awirb (changes requested):

PR #526, opened 2021-08-25
add more and update pytests

awirb:


avaframe/com1DFA/com1DFA.py
```diff
@@ -805,10 +807,17 @@ def placeParticles(massCell, indx, indy, csz, massPerPart, rng, initPartDistType
         # place particles randomly in the cell
         xpart = csz * (rng.random(nPart) - 0.5 + indx)
         ypart = csz * (rng.random(nPart) - 0.5 + indy)
+    elif initPartDistType == 'uniform':
+        # place particles equaly distributed
+        xpart = csz * (- 0.5 + indx) + x
+        ypart = csz * (- 0.5 + indy) + y
     else:
+        log.warning('Chosen value for initial particle distribution type not available: %s uniform is used instead' %
+                    initPartDistType)
```
fso42, line 816: I suggest combining these two, and add an `if initPartDistType is not uniform` -> log warning

fso42 (changes requested):

avaframe/com1DFA/com1DFA.py
awirb: done

PR #536, opened 2021-09-01
fix error in explicit friction force and add pytests for cython

matthiasto:


avaframe/tests/test_DFAfunctionsCython.py
```diff
@@ -235,7 +235,7 @@ def test_reprojectionC(capfd):
 
     for reprojectionIterations in range(10):
         xpn, ypn, iCell, Lx0, Ly0, w0, w1, w2, w3 = DFAfunC.samosProjectionIteratrive(x1, y1, z1,
-                        ZZ, Nx, Ny, Nz, csz, ncols, nrows, interpOption, reprojectionIterations)
+                                                                                      ZZ, Nx, Ny, Nz, csz, ncols, nrows, interpOption, reprojectionIterations)
```
fso42, line 238: For this (and the line lower down) I'd prefer to rather violate the indent pep but not have such long lines...

fso42 (changes requested):

avaframe/tests/test_DFAfunctionsCython.py
matthiasto: atom did this automatically, I ll revert it

PR #542, opened 2021-09-08
update aimec for using DEM with different spatial resolution

awirb:


fso42 (approved):

docs/moduleAna3AIMEC.rst
```diff
@@ -25,6 +25,11 @@ Inputs
 *  Results from avalanche simulation (when using results from com1DFA,
    the helper function :py:func:`ana3AIMEC.dfa2Aimec.mainDfa2Aimec` in :py:mod:`ana3AIMEC.dfa2Aimec` fetches and prepares the input for Aimec)
 
+.. Note:: If the results have been performed by remeshing the DEM to a different spatial resolution, it is still
+          possible to provide a DEM of a different spatial resolution compared to the result files to aimec.
+          However, the z coordinates used to compute the location of the runout point and the fall height (deltaH)
+          are then derived from the provided dem data.
```
matthiasto, line 31: I would write:
The DEM resolution and extend can be different from the result raster ones (but all result raster extend and resolution need to be identical). The DEM file is required in to compute the z coordinate of the avalanche path as well as to compute de fall height (deltaH) of the avalanche.

avaframe/runScripts/runComparisonModules.py
```diff
@@ -134,6 +136,7 @@
             else:
                 cfgAimec['FLAGS']['flagMass'] = 'False'
             cfgAimec['AIMECSETUP']['comModules'] = 'com1DFAOrig|com1DFA'
+            cfgAimec['AIMECSETUP']['resType'] = 'ppr'
```
matthiasto, line 140: why do you hardcode this? why not reed the ini file?

fso42 (approved):

matthiasto (changes requested):
I havent tried to run it, will try when I am finished with my PR

avaframe/runScripts/runComparisonModules.py
awirb: this if for the report generation of the comparison report - where we use the locals for com1DFA and com1DFAOrig for this runScript to perform the ava simulations, but to run aimec and generate the report we use some fixed settings in the  runScripts - that is for this runScript one doesn't have to change the aimec configuration - but it is set how we defined it for this test

docs/moduleAna3AIMEC.rst
awirb: I updated this sentence

docs/moduleAna3AIMEC.rst
```diff
@@ -25,6 +25,12 @@ Inputs
 *  Results from avalanche simulation (when using results from com1DFA,
    the helper function :py:func:`ana3AIMEC.dfa2Aimec.mainDfa2Aimec` in :py:mod:`ana3AIMEC.dfa2Aimec` fetches and prepares the input for Aimec)
 
+.. Note:: The spatial resolution of the DEM and its extend can differ from the result raster data.
+          However, all the result raster have to be of identical spatial resolution and extent.
+          This might be of interest if the result rasters have been derived by remeshing the DEM to different spatial
+          resolution. The DEM data is used to compute the z coordinate of the location of the runout point as well as
+          the fall height (deltaH).
+
```
matthiasto, line 33: I like this

avaframe/runScripts/runComparisonModules.py
matthiasto: mm still not convinced

fso42 (approved):

PR #549, opened 2021-09-14
remove runCom1DFA function inside and call com1DFAMain

awirb:
* update runScripts
* add pytest


avaframe/com1DFA/com1DFA.py
```diff
@@ -37,7 +43,106 @@
...
+    modCfg, variationDict = dP.getParameterVariationInfo(avalancheDir, com1DFA, cfgFile, variationDict)
+
+    # fetch input data - dem, release-, entrainment- and resistance areas
+    inputSimFiles = gI.getInputDataCom1DFA(avalancheDir, modCfg['FLAGS'])
+
+    # write full configuration file to file
+    cfgUtils.writeCfgFile(avalancheDir, com1DFA, modCfg, fileName='sourceConfiguration')
+
+    # create a list of simulations and generate an individual configuration object for each simulation
+    # if need to reproduce exactely the hash - need to be strings with exactely the same number of digits!!
```
fso42, line 88: Typo: exactly

avaframe/in3Utils/cfgUtils.py
```diff
@@ -102,10 +102,14 @@ def getModuleConfig(module, fileOverride='', modInfo=False):
...
-
-    return cfg
+    if modInfo:
+        # Finally read it
+        cfg, modDict = compareConfig(iniFile, modName, compare, modInfo)
+        return cfg, modDict
+    else:
+        # Finally read it
+        cfg = compareConfig(iniFile, modName, compare, modInfo)
+        return cfg
```
fso42, line 112: I suggest: 
```suggestion
    # Finally read it
    cfg, modDict = compareConfig(iniFile, modName, compare, modInfo)
    if modInfo:
        return cfg, modDict
    return cfg
```

avaframe/runScripts/runTestFP.py
```diff
@@ -10,7 +10,6 @@
 import avaframe.in3Utils.fileHandlerUtils as fU
 from avaframe.in1Data import getInput as gI
 import avaframe.com1DFA.com1DFA as com1DFA
-from avaframe import runCom1DFA
 import avaframe.ana1Tests.FPtest as FPtest
```
fso42, line 13: This script fails with:
`ModuleNotFoundError: No module named 'avaframe.out3Plot.makePalette'`

and the com1DFA import is also not the correct one, right?

fso42 (changes requested):

avaframe/com1DFA/com1DFA.py
awirb: done

avaframe/in3Utils/cfgUtils.py
awirb: this is not possible as compareConfig only returns the modDict if modInfo is true

avaframe/runScripts/runTestFP.py
awirb: I remove the import as it is not required and changed the import of com1DFA, also removed the hSPH as it is not added to the particles anymore - but still fails -made an issue out of it

avaframe/in3Utils/cfgUtils.py
matthiasto: you could just always return the modDict. This way felixs solution works and is shorter and nicer and if you call this function somewhere else and don't want to use the modDict just do:
cfg, _ = compareConfig(iniFile, modName, compare, modInfo)

matthiasto (changes requested):

avaframe/in3Utils/cfgUtils.py
awirb: changed now modDict is always returned and updated the calls where it is not required

matthiasto (approved):

fso42 (approved):

PR #552, opened 2021-09-15
fix testFP

matthiasto:
* fix issue #551


awirb (commented):
suggestion: make figure wider

PR #553, opened 2021-09-15
clean up runScripts using with respect to ini files and directly setting parameters

awirb:
* move setting of parameters within runScript to top
* clean up naming of test report directories,...
* remove aimec ini files if no parameter specific for one ava is different from default or settings set in runScript
* update runStandardTests and runVariations - both still only available for com1DFAOrig


fso42 (approved):

matthiasto (approved):
Ha I like this, now its clear where we have to look for the inputs!
Also good that you cleaned the benchmark ini files.
Do I need to test something?

Issue #548, opened 2021-09-14
QGis run fails on com2AB


fso42:
com2AB fails for a certain data setup (cannot publish it due to license issues; available upon request), see error message below

Version avaframe: 0.5.12, QGis connector 0.9, QGis 3.12 / 3.20

![image](https://user-images.githubusercontent.com/6945681/133227203-2523c08c-af55-4f68-ae23-d965abd2a47b.png)

(reported by C.T. WLV)

fso42:
Reason is a missing record 
```
com2AB - INFO -  Running Alpha Beta 4.1 on: CTAVA 
outAB - WARNING -  -2 SD out of profile
```
fix by checkin of number of records

PR #554, opened 2021-09-15
Deploy to pypi 

fso42:
- Add setup and workflow to deploy to pypi (manual workflow start for now)
- new way to compile the cython code:
  ```python setup.py build_ext --inplace```
-  remove codecoverage (codeclimate will be used)
- complete rework of installation instructions; doc build for this branch https://docs.avaframe.org/en/pypi/
- bug fix for com2AB #548


avaframe/tests/test_com2AB.py
```diff
@@ -108,6 +108,24 @@ def test_calcAB(capfd):
...
+    '''test writing to shapefile'''
+
+    avaName = 'avaSlide'
+    dirname = pathlib.Path(__file__).parents[0]
+    avalancheDir = dirname / '..' / 'data' / avaName
+    cfg = cfgUtils.getModuleConfig(com2AB)
+
+    # run main routine
+    resAB = com2AB.com2ABMain(cfg, avalancheDir)
+    abShpFile = outAB.writeABtoSHP(resAB)
```
matthiasto, line 121: I think Anna said we should first copy the inputs to a temporary path, run and save the results there so that it is then automatically removed at the end of the test.

fso42: ah, totally forgot! Thanks, will fix

docs/installation.rst
```diff
@@ -36,31 +112,27 @@ Setup AvaFrame
...
-   :literal:
+Compile the cython com1DFA part. You might also have to install a c-compiler (gcc or similar) through your systems
+package manager::
 
-Install these with (conda-forge is needed for cmcrameri)::
+  python setup.py build_ext --inplace
 
-  conda install -c conda-forge --file requirements.txt
+.. Warning::
+   You will have to do this compilation every time something changes in the cython code.
```
awirb, line 133: suggestion: add a hint that it is a good idea to do this every time changes are pulled from the master

awirb (commented):
A lot more examples can be found in the ``runScripts`` directory. - maybe it would be nice to reference where they area listed in the Develop chapter - for an overview of available runScripts

Issue #550, opened 2021-09-14
add runStandardTests for com1DFA


awirb:
so far runStandardTests only for com1DFAOrig

awirb:
currently the runScript for running the standard test is only available for com1DFAOrig, also for the variations tests 
- update to com1DFA as well

awirb:
now there are runScripts for com1DFA too

PR #556, opened 2021-09-17
add standard and variations test scripts for com1DFA

awirb:
* add run scripts for standard tests and variations tests com1DFA
* add ini files required 
* switch to pathlib 
* address issue #550

### DIFF
--- a/avaframe/runScripts/runStandardTestsCom1DFAOrig.py
+++ b/avaframe/runScripts/runStandardTestsCom1DFAOrig.py
@@ -22,7 +22,7 @@ from avaframe.in3Utils import cfgUtils
 from avaframe.in3Utils import logUtils
 from benchmarks import simParametersDict
 
-#+++++++++REQUIRED+++++++++++++
+# +++++++++REQUIRED+++++++++++++
 # Which result types for comparison plots
 outputVariable = ['ppr', 'pfd', 'pfv']
 # aimec settings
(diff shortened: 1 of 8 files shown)


avaframe/runScripts/runStandardTestsCom1DFAOrig.py
```diff
@@ -87,8 +85,8 @@
 
     # Load input parameters from configuration file for standard tests
     # write config to log file
-    avaName = os.path.basename(avaDir)
-    standardCfg = os.path.join('..', 'benchmarks', test['NAME'], '%s_com1DFAOrigCfg.ini' % test['AVANAME'])
+    avaName = pathlib.Path(avaDir).name
```
fso42, line 88: Is this actually needed, since the AvaName also lives in test['AVANAME']?

avaframe/runScripts/runStandardTestsCom1DFAOrig.py
```diff
@@ -108,14 +106,19 @@
     plotDict = oP.plotAllPeakFields(avaDir, cfgMain['FLAGS'], modName)
 
     # Set directory for report
-    reportDir = os.path.join(avaDir, 'Outputs', modName, 'reports')
+    reportDir = pathlib.Path(avaDir, 'Outputs', modName, 'reports')
```
fso42, line 109: Would it make sense to convert avaDir to a pathlib object right at the beginning so that most of these lines can be converted to `avaDir / Outputs / ....` ?

fso42 (approved):

avaframe/runScripts/runStandardTestsCom1DFAOrig.py
awirb: removed

Issue #494, opened 2021-07-08
add d0 to real ava release shape files


awirb:
no d0 provided except for avaGar

awirb:
added default value of thickness=1.0 for release areas to shape files if nothing else was provided - also switched from d0 to thickness

PR #557, opened 2021-09-17
Add option to force reading relTh from ini 

awirb:
* update log so that it tells you where the release thickness is read from in addition to its value 
* update docu with a note on this 
* update glossary - more info required?
* update shapefiles 
* update warning
* also read entTh from shape file 
* hEnt -> entTh 
* close issue #494  here


fso42 (approved):

avaframe/in2Trans/shpConversion.py
```diff
@@ -57,11 +57,11 @@ def SHP2Array(infile, defname=None):
     infile = pathlib.Path(infile)
     # set defaults for variables
     layername = None
-    d0 = None
+    relTh = None
     rho = None
     sks = None
     iso = None
-    d01 = []
+    relTh1 = []
```
fso42, line 64: What is this needed for?

docs/moduleCom1DFA.rst
```diff
@@ -43,6 +43,9 @@ In the directory ``Inputs``, the following files are required:
 * release area scenario as shapefile (in Inputs/REL); multiple features are possible
   -> the release area name should not contain an underscore, if so '_AF' is added
 
+.. Note:: Release thickness can be specified in the shape file by adding an attribute called `relTh` or via the
```
fso42, line 46: Please add:
For backwards compatibility, the attribute 'd0' also works, but we suggest to use relTh in new projects.

fso42 (changes requested):

avaframe/in2Trans/shpConversion.py
awirb: this is the list where the thickness values of the individual features are appended i renamed it to thicknessList and moved it to the respective list initialisation for the name

docs/moduleCom1DFA.rst
awirb: Added

avaframe/com1DFA/com1DFA.py
```diff
@@ -220,7 +223,7 @@ def com1DFACore(cfg, avaDir, cuSimName, inputSimFiles, outDir, relThField):
     return particlesList, fieldsList, Tsave, dem, reportDict, cfg
 
 
-def prepareRelase(cfg, rel, inputSimLines):
+def prepareReleaseEntrainment(cfg, rel, inputSimLines):
```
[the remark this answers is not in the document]
fso42: inputSimLines is missing from the docstring

avaframe/com1DFA/com1DFA.py
```diff
@@ -253,14 +256,14 @@ def prepareRelase(cfg, rel, inputSimLines):
...
+    releaseLine['thicknessSource'] = [''] * len(releaseLine['thickness'])
+    for k in range(len(releaseLine['thickness'])):
+        if cfg['GENERAL'].getboolean('useRelThFromIni') or releaseLine['thickness'][k] == 'None':
+            releaseLine['thickness'][k] = cfg['GENERAL'].getfloat('relTh')
+            releaseLine['thicknessSource'][k] = 'ini file'
         else:
-            releaseLine['relTh'][k] = float(releaseLine['relTh'][k])
-            releaseLine['relThSource'][k] = 'shp file'
+            releaseLine['thickness'][k] = float(releaseLine['thickness'][k])
+            releaseLine['thicknessSource'][k] = 'shp file'
```
fso42: This (and the other similiar ones) can be simplified (at least I think so):
- if use from ini is set, all release lines get the same value, ie. all thicknesses and source can be set at once
- if shp file is set, all sources can be set at once,
- therefore the only loop that is needed is that for the conversion to float

fso42 (changes requested):

docs/glossary.rst
```diff
@@ -14,13 +14,13 @@ Glossary
...
+      consistently below 10 degrees.
 
    beta angle
-      Angle from the topmost point of a profile to the :term:`beta point`. 
+      Angle from the topmost point of a profile to the :term:`beta point`.
 
-   release thickness [*relTh*, *d0*]
-      TODO: which symbol 
+   release thickness [*relTh*]
+      TODO: which symbol
```
fso42, line 23:
```suggestion
   release thickness [*relTh*]
      ```

Also add the same for entrainment thickness

Comment (awirb) [reply docs/glossary.rst:23]: done

File: avaframe/com1DFA/com1DFA.py [hunk 6]
```diff
@@ -244,20 +250,17 @@ def prepareRelase(cfg, rel, inputSimLines):
...
         the suffix _AF will be added for the simulation name')
-    releaseLine = inputSimLines['releaseLine']
-    for k in range(len(releaseLine['d0'])):
-        if releaseLine['d0'][k] == 'None':
-            releaseLine['d0'][k] = cfg['GENERAL'].getfloat('relTh')
-        else:
-            releaseLine['d0'][k] = float(releaseLine['d0'][k])
+
+    # set release thickness
+    releaseLine = setThickness(cfg, inputSimLines['releaseLine'], 'useRelThFromIni', 'relTh')
```
fso42, line 263: useRelThFromIni is hardcoded now? Is that wanted?

avaframe/com1DFA/com1DFA.py
```diff
@@ -266,21 +269,50 @@ def prepareRelase(cfg, rel, inputSimLines):
...
 
 
+def setThickness(cfg, lineTh, flagTh, typeTh):
+    """ set thickness in line dictionary for release area, entrainment area
+
+    Parameters
+    -----------
+    lineTh: dict
+        dictionary with info on line (e.g. release area line)
+    flagTh: bool
```
fso42, line 293: See issue #86 : we are trying not to use flagSOMETHING, so maybe change this to useThFromIni

awirb: changed to useThFromIni

avaframe/com1DFA/com1DFA.py
awirb: yes, this is the flag that is passed to the function setThickness (which is new - to reduce duplication of code as was almost the same code block for release, secrel and entrainment)

PR #572, opened 2021-10-04
update doc strings

matthiasto:


matthiasto:
Just an example for Amaury

Issue #588, opened 2021-11-16
Pytests fail 


fso42:
Fails on 3.9 python with updated modules. I.e. conda environment with 3.9.X and `conda upgrade --all within this environment

- [x]  some due to a suspected matplotlib update -> maybe fixed in `fixAction` branch, function `makeColorMap`

  - @matthiasto: please check whether this routine still gives the results as intended
- [x] test_outQuickPlot.py fails due to an assertion error: @awirb please look into this
- [ ] There is a warning:
`out3Plot/outAIMEC.py:581: MatplotlibDeprecationWarning: You are modifying the state of a globally registered colormap. This has been deprecated since 3.3 and in 3.6, you will not be able to modify a registered colormap in-place. To remove this warning, you can make a copy of the colormap first. cmap = mpl.cm.get_cmap("broc").copy()
cmap.set_bad(color='w')
`
which should be adressed: @matthiasto please look into this
- [x] `test_ana3Aimec` fails with some error about threshold. @matthiasto plese look into this

PR #591, opened 2021-11-17
Fix action

matthiasto:
- Fix the pytest. It seems it was related to some new behavior of matplotlib colors and legend.
- update the pytest github action to version 3.9


fso42 (changes requested):
I suggest we address the warning in issue #588 right away to avoid running into a similar problem soon...

fso42 (approved):

Issue #627, opened 2021-12-30
Move relThField to Ini file


fso42:
Similiar problem to #576 : if this is set via function call, the hash and info is lost afterwards
Solution:
require the field to be written to file in beforehand, then supply it via a flag `useRelThField` and `relThFieldFile` parameter

awirb:
now it is using relThFromFile flag if True release thickness is read from an asc file in avaDir/Inputs/RELTH

PR #628, opened 2022-01-03
Add setting of thickness workflow

awirb:
* thFromShp flags for release thickness, entrainment thickness and second. release thickness 
* if read from shp - check if thickness values provided for all features and add info to ini file 

this enables a unique simHash including info on release scenario with correct thickness - hence also possible to avoid duplicate runs of identical sims 

parameter variation still possible but more options:
* for thFromShp = False, relTh is used for all features of a a shp file - variation possible using start:end:steps, value1|value2|value3|... 
* for thFromShp= False or True - thPercentVariation - for example vary relTh in +- 50 percentage of refValue in 10 steps in format relTh= 50$10, also possible just positive variation +50$10, or only negative -50$10
* fix issue #578 
TODO: 

filter option for relTh - done - however it is com1DFA specific... how to integrate this properly? with modName as input?
address issue #627


avaframe/ana4Stats/probAna.py
```diff
@@ -148,55 +165,63 @@ def probAnalysis(avaDir, cfg, module, parametersDict='', inputDir=''):
...
-
-    # # Save to .asc file
-    avaName = avaDir.name
-    outFileName = '%s_probMap%s.asc' % (avaName, cfg['GENERAL']['peakLim'])
-    outFile = outDir / outFileName
-    IOf.writeResultToAsc(header, probMap, outFile)
+
+    # initialize flag if analysis has been performed or e.g. no matching files found
+    analysisPerformed = False
+    if simNameList != []:
```
fso42, line 171: The is no path in here if simListName is empty, and the rest is indented just because of this. 
So I suggest adding a second return here with a simple 
if simNamelinst is empty
  analysisPerformed = False
  return ana....
And then don't indent the rest....

avaframe/ana4Stats/probAna.py
```diff
@@ -93,30 +92,48 @@ def updateCfgRange(cfg1, cfgProb, varName):
...
 
-    # if reference value is not in this list - add reference values
-    valValues = np.linspace(valVal-valRange, valVal+valRange, valSteps)
-    refIn = False
-    if valVal not in valValues:
-        cfg1['GENERAL'][varName] = '%s:%s:%s&%s' % (str(valVal-valRange), str(valVal+valRange), str(valSteps), str(valVal))
-        refIn = True
+        # if reference value is not in variation - add reference values
+        if valVal not in valValues:
+            cfg1['GENERAL'][varName] = cfg1['GENERAL'][varName] + '&' + str(valVal)
```
fso42, line 134: Break this out to an extra function?

avaframe/com1DFA/com1DFA.py
```diff
@@ -397,16 +471,23 @@ def prepareInputData(inputSimFiles):
...
+            secondaryReleaseLine['fileName'] = [secondaryReleaseFile]
+            secondaryReleaseLine['type'] = 'Secondary release'
+        else:
+            message = 'No secondary release file found'
+            log.error(message)
+            raise FileNotFoundError(message)
     else:
         secondaryReleaseLine = None
+        # set False
+        entResInfo['flagSecondaryRelease'] = 'No'
```
fso42, line 487: I suggest using `False` insted of `'No'`

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -41,14 +40,23 @@ def getVariationDict(avaDir, fullCfg, modDict):
...
-                locValue = fU.splitIniValueToArraySteps(value)
+            if key in ['relThPercentVariation', 'entThPercentVariation', 'secondaryRelThPercentVariation'] and value != '':
+                # here the factor for changing thValues is added to the variationDict instead of the
+                # values directly
+                locValue = splitVariationToArraySteps(value, key, fullCfg)
                 variations[key] = locValue
                 defValue = modDict[section][key][1]
                 log.info('%s: %s (default value was: %s)' % (key, locValue, defValue))
+            else:
+                if ':' in value or '|' in value or '$' in value:
```
fso42, line 55: if value in [':','|' ...
?

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -197,3 +194,245 @@ def checkRelEntThVariation(cfg, variationDict):
...
+        # if thickness should be read from ini file - check if format is correct
+        if '$' in cfg['GENERAL'][thType] and len(cfg['GENERAL'][thType].split('$')) != 3:
+            message = 'Format of relTh value in ini file is not correct - for variation from ini use refValue$percent$numberOfSteps'
+            log.error(message)
+            raise AssertionError(message)
+
+    return cfg
+
+
+def checkThicknessSettings(cfg, thName):
```
fso42, line 259: Maybe return a boolean like `thicknessIsCorrect` os similar so it can be used in the calling function

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -197,3 +194,245 @@ def checkRelEntThVariation(cfg, variationDict):
...
+                raise AssertionError(message)
+
+    else:
+        message = 'Check %s - needs to be True or False' % (thFlag)
+        log.error(message)
+        raise AssertionError(message)
+
+
+def splitVariationToArraySteps(value, key, fullCfg):
+    """ split variation in percent to create a list of factors to set paramerter value for variations
```
fso42, line 294:
```suggestion
    """ split variation in percent to create a list of factors to set parameter value for variations
```

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -197,3 +194,245 @@ def checkRelEntThVariation(cfg, variationDict):
...
+
+    # only add entries to cfg if appropriate for chosen simType (e.g. entTh if ent or entres run)
+    entCondition = (key == 'entThPercentVariation' and 'ent' in simType)
+    secRelCondition = (key == 'secondaryRelThPercentVariation' and cfg['GENERAL']['secRelArea'] == 'True')
+    relCondition = (key == 'relThPercentVariation')
+
+    # fetch variation factor
+    variationFactor = float(row._asdict()[key])
+
+    # update thickness values accoridng to variation
```
fso42, line 366:
```suggestion
    # update thickness values according to variation
```

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -197,3 +194,245 @@ def checkRelEntThVariation(cfg, variationDict):
...
+            for count, id in enumerate(idList):
+                thNameId = thType + id
+                cfg['GENERAL'][thNameId] = str(float(thicknessList[count]) * variationFactor)
+
+            # set percentVaration parameter to actual variation in percent$steps
+            if variationFactor == 1.:
+                # if variation is 0% set percentVaration = ''
+                variationIni = ''
+            elif variationFactor < 1:
+                variationIni = '-' + str((1. - variationFactor) * 100) + '$1'
```
fso42, line 386: Am I correct that the lowest allowed percentVariation in % is then 1%? Do we need to mention this?

avaframe/in3Utils/cfgUtils.py
```diff
@@ -202,7 +202,19 @@ def compareConfig(iniFile, modName, compare, modInfo=False, toPrint=True):
         for section in locCfg.sections():
             if defCfg.has_section(section):
                 for key in locCfg.items(section):
-                    log.warning('Additional Key [\'%s\'] in section [\'%s\'] is ignored.' % (key[0], section))
+                    # an exception is made for thickness values that are added for the features of a releaseScenario,
+                    # entrainment Scenario or secondar. release scenario
+                    # these are added to the configuration and also to the modDict if variation is applied
+                    validItems = ['entrainmentScenario', 'DEM', 'secondaryReleaseScenario']
+                    if 'relTh' in key[0] or 'entTh' in key[0] or 'secondaryRelTh' in key[0] or key[0] in validItems:
```
fso42, line 209: rewrite to if `key[0] in ...` ?

avaframe/in3Utils/cfgUtils.py
```diff
@@ -202,7 +202,19 @@ def compareConfig(iniFile, modName, compare, modInfo=False, toPrint=True):
...
+                        locValue = locCfg.get(section, key[0])
+                        cfg.set(section, key[0], locValue)
+                        log.info('\t\t%s : %s added to %s' % (key[0], locValue, section))
```
fso42, line 212: this is internal, so I would set it to debug

avaframe/in3Utils/cfgUtils.py
```diff
@@ -668,3 +691,75 @@ def orderSimFiles(avalancheDir, inputDir, varParList, ascendingOrder, specDir=''
...
+        notIn = True
+
+    # create required parameters for searching
+    thFlag = key + 'FromShp'
+    thId = key + 'Id'
+    thThickness = key + 'Thickness'
+    thPercentVariation = key + 'PercentVariation'
+
+    # append identifier if simulation matches thickness filter criteria
+    simDF['toBeAdded'] = 'False'
```
fso42, line 730: Does this need to be a string?

fso42 (changes requested):

avaframe/ana4Stats/probAna.py
awirb: done

avaframe/ana4Stats/probAna.py
```diff
@@ -83,8 +84,6 @@ def updateCfgRange(cfg1, cfgProb, varName):
         --------
         cfg1: configParser
```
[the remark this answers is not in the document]
awirb: changed to cfg

avaframe/com1DFA/deriveParameterSet.py
awirb: I think this won't work as the it should check if in the value ':' is contained not if the value is ':'

avaframe/com1DFA/deriveParameterSet.py
awirb: added a flag thicknessSettingsCorrect

avaframe/com1DFA/deriveParameterSet.py
awirb: the percent variation can be anything between 0 and 1

awirb: the 1 here corresponds to the variationFactor so 1 means value x1 - no variation or 0% variation - if the factor is 0.5 - 50% percent variation, ..

avaframe/in3Utils/cfgUtils.py
awirb: here again it should also find e.g. relTh0 so we need to check if 'relTh' in key[0]

avaframe/in3Utils/cfgUtils.py
awirb: changed to bool

avaframe/ana4Stats/probAna.py
awirb: I would suggest to keep this in one because this second part here just sets the actual values after first checking if no invalid settings are used and building the parameters

avaframe/com1DFA/com1DFA.py
awirb: these 'flags' are set to No and Yes for also entrainment and resistance as they are then directly stated in the markdown report - and initially chosen to be more readable as yes and no instead of true and false - should we keep this or change it throughout? but this has an effect on many places so if we want to change it I suggest making an issue

fso42: Didn't realize the report connection, keep it

avaframe/com1DFA/deriveParameterSet.py
fso42: `if any(c in value for c in [':', '$', '|'])` then?

avaframe/in3Utils/cfgUtils.py
fso42: Same as before:
`searchKeys = ['relTh','entTh'....]
if any(k in key[0] for k in searchKeys)
`?

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -197,3 +194,254 @@ def checkRelEntThVariation(cfg, variationDict):
...
+
+    else:
+        message = 'Check %s - needs to be True or False' % (thFlag)
+        log.error(message)
+        raise AssertionError(message)
+
+    # if no error ocurred - thickness settings are correct
+    thicknessSettingsCorrect = True
+
+    return thicknessSettingsCorrect
```
matthiasto, line 300: why do you return something? it fails anyways if there is a problem

avaframe/tests/data/avaFilterTest/com1DFA/configurationFiles/relGar_ent_dfa_789ce37489.ini
```diff
@@ -0,0 +1,119 @@
...
+seed = 12345
+iniStep = False
+subgridMixingFactorIni = 10.
+dtIni = 0.1
+sphOptionIni = 2
+stopCritIni = 0.1
+stopCritIniSmall = 1.001
+maxIterations = 100
+bufferZoneFactor = 4
+additionallyFixedFactor = 0.275
```
matthiasto, line 18: Didn't we say we only put parameters that change from the default cfg in here?

matthiasto (changes requested):
- The configuration is printed 3 time at the beginning of the simulation, maybe this could be reduced to only one no?
- If we want a null simulation, why do we have to go and change the settings for entrainment? (say if we want entrainment from shp or ini file)
-

avaframe/com1DFA/com1DFA.py
```diff
@@ -80,88 +110,119 @@ def com1DFAMain(avalancheDir, cfgMain, cfgFile='', relThField='', variationDict=
...
-        # append new simulations configuration to old ones (if they exist), return  total dataFrame and write it to csv
-        simDFNew = simDF.append(simDFOld)
-        cfgUtils.writeAllConfigurationInfo(avalancheDir, simDFNew, specDir='')
-
-        # write full configuration (.ini file) to file
-        date = datetime.today()
-        fileName = 'sourceConfiguration_' + '{:%d_%m_%Y_%H_%M_%S}'.format(date)
-        cfgUtils.writeCfgFile(avalancheDir, com1DFA, modCfg, fileName=fileName)
-        return particlesList, fieldsList, tSave, dem, plotDict, reportDictList, simDF
+        return dem, plotDict, reportDictList, simDFNew
```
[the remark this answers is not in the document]
awirb: com1DFACore returns them but for com1DFAMain the problem is, this only refers to the last simulation performed so I though that is not that useful

avaframe/com1DFA/com1DFA.py
```diff
@@ -244,7 +305,7 @@ def com1DFACore(cfg, avaDir, cuSimName, inputSimFiles, outDir, relThField=''):
     # add time and mass info to report
     reportDict = reportAddTimeMassInfo(reportDict, tCPUDFA, infoDict)
 
-    return particlesList, fieldsList, Tsave, dem, reportDict, cfg, infoDict['tCPU']
+    return particlesList, fieldsList, Tsave, dem, reportDict, cfg, infoDict['tCPU'], inputSimFiles
```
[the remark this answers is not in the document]
awirb: I think we could leave the outputs here in case one wants it when just running the com1DFAcore function

avaframe/tests/data/avaFilterTest/com1DFA/configurationFiles/relGar_ent_dfa_789ce37489.ini
awirb: but this is not an input for running com1DFA this is used to test the filtering  functions that read the ini files from the given input folder or from avaDir/Outputs/com1DFA/configurationFiles so here i need the full ini

avaframe/com1DFA/com1DFA.py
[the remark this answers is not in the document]
fso42: I'm fine with removing it, but if you want to keep it, move it to the end of the line. If it is too much effort for this PR -> new issue

avaframe/com1DFA/deriveParameterSet.py
awirb: changed

avaframe/com1DFA/deriveParameterSet.py
fso42: I think it makes sense to return True in case you want to use it in the calling function. Otherwise the return value is None... (https://realpython.com/python-return-statement/#implicit-return-statements)

avaframe/in3Utils/cfgUtils.py
awirb: validItems = ['entrainmentScenario', 'DEM', 'secondaryReleaseScenario']
searchItems = ['relTh', 'entTh', 'secondaryRelTh']
if any(s in key[0] for s in searchItems) or key[0] in validItems:

avaframe/com1DFA/com1DFA.py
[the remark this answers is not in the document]
matthiasto: Ok so maybe we should not return it

avaframe/com1DFA/deriveParameterSet.py
matthiasto: ok but then do we need to catch the value when we call the function?

fso42: no, not necessary. But possible if needed :-)

awirb:
If you chose null sim - no error if ent files is available but has no thickness attritubte

avaframe/com1DFA/deriveParameterSet.py
awirb: so I leave it like it is now?

matthiasto (approved):

fso42 (approved):

avaframe/com1DFA/com1DFA.py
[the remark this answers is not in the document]
awirb: moved particlesList, fieldsList, Tsave to the end of the returns

PR #629, opened 2022-01-04
Ana1 influence test

belieresfrendo:
Hi, this is a draft branch
-> saving and plotting results isn't finished yet


fso42 (changes requested):
Thanks Amaury for your  PR. 

Before I look more detailed into the code, a few general remarks that you could address right away:

- [ ] add a description of what the code in the PR does / genereal idea behind it -> so it is easier to understand what you are trying to do
- [ ] drop the ana Part from your filenames in ana1Test -> easier to read and shorter, since it is in the ana1 module, its aim is clear.
- [ ] remove the folder `anaInfluenceTest`. If you need the files therein, move them to ana1 and name them like InfluenceTestRefCfg.ini or similar -> less clutter in the main directory
- [ ] do not make the core DFA functions specific for your application. All the checking needs to be done inside your code (module or runscript).  -> we cannot maintain a second version of the core function for a specific task...
- [ ] avoid `print` and use log.info/debug/warning as needed
- [ ] split out the plotting to an `out` module and add the plotting to the runscript instead
- [ ] remove comments/ code fragments that you don't need anymore 
- [ ] address the pep8 issues

PR #637, opened 2022-01-13
update runScripts with new return from com1DFAMain

awirb:
* now only 4 return parameters from com1DFAMain - update of runScripts was required 
* update benchmark ini files releaseScenario (move to section INPUT) and relTh (add relThFromShp flag value) values according to new settings


avaframe/com1DFA/com1DFATools.py
```diff
@@ -228,3 +230,37 @@ def extendProfileBottom(cfg, dem, profile):
...
+        inputDir = pathlib.Path(inputDir)
+
+    for tStep in timeSteps:
+        tFiles = sorted(list(inputDir.glob('*%s_t%.2f*' % (resType, tStep))))
+
+    fieldsList = []
+    for tFile in tFiles:
+        fieldsList.append({resType: IOf.readRaster(str(tFile))['rasterData']})
+
+    return fieldsList
```
matthiasto, line 266: there is in com1DFA a readFields function, could they maybe be combined or is there a difference between the two?

matthiasto (changes requested):

avaframe/com1DFA/com1DFATools.py
awirb: I merged the two functions

avaframe/com1DFA/com1DFA.py
```diff
@@ -1888,20 +1891,25 @@ def trackParticles(cfgTrackPart, dem, particlesList):
...
+        resType: list
+            list of desired result types
         simName : str
             simulation name
         flagAvaDir: bool
             if True inDir corresponds to an avalanche directory and pickles are
             read from avaDir/Outputs/com1DFA/particles
         comModule: str
             module that computed the particles
+        timeStep: float
```
fso42, line 1910: How does this have to be specified? Does it have to be precise or is a range considered? -> Provide info here

avaframe/com1DFA/particleTools.py
```diff
@@ -21,20 +21,22 @@
 log = logging.getLogger(__name__)
 
 
-def initialiseParticlesFromFile(cfg, avaDir):
+def initialiseParticlesFromFile(cfg, avaDir, releaseScenario):
     # TODO: this is for development purposes, change or remove in the future
     # If initialisation from file
+
+    print('RELEASE ', releaseScenario)
```
fso42, line 28: Thi seems to be a debut print? If not, convert to log...

fso42 (changes requested):

avaframe/com1DFA/particleTools.py
awirb: removed print

avaframe/com1DFA/com1DFA.py
awirb: I added an atol of 1.e-6 as the file name where the time info is read if saved from com1DFA is set to %.2f so could even be less

matthiasto (approved):

fso42 (approved):

PR #638, opened 2022-01-14
Ana1 influence test

belieresfrendo:
This module developped in ana1Tests aims to run simulations and compare them.
The comparison is made according 2 parameters.
the first one will be an axis parameter, while the second one will allow to print several curves on the same graphic to see how the evolve according the first parameter.

-------------------------------
Moreover, some points are improved in this PR.
-> New script in DFA module to check if .ini files are well parametered
-> modify DFA to be able to get easily sim Hashes which has to be performed
-> change one spelling fault in a file name (initialise -> initialize)


docs/DFAnumerics.rst
```diff
@@ -544,6 +547,37 @@ mean mesh velocity and then the implicit term which leads to:
...
+where :math:`\mathbf{u}_{ij} = \mathbf{u}_i - \mathbf{u}_j` is the relative velocity
+between particle i and j, and :math:`\lambda_{ij} = \frac{c_i+c_j}{2}` with
+:math:`c_i = \sqrt{gh_i}`, the wave speed. The :math:`\lambda_{ij}` is obtained as
+turning expressions related to time and spacial discretization parameters
+into an expression on maximal speed between both particles in the Lax Friedrich
+scheme.
+
+CFL condition
+"""""""""""""
+
```
matthiasto, line 578: I am not sure where we add this section. Remove it for now, I will take care of adding this when we get something meaningfull

matthiasto (changes requested):
Your branch is not up to date with the master. Please rebase it.
I also see that the com1DFA.py and DFAFunctions.pyx have weird changes. I think you need to replace them by the ones from the master to avoid a lot of conflicts (also pay attention that your atom editor does not remove or add blank lines on its own)

fso42:
Closing in favour of #661

PR #647, opened 2022-01-24
Add range option

awirb:
* add in addition to vary thickness values if read from shp - not just in percent but also in absolute values

### DIFF
--- a/avaframe/com1DFA/deriveParameterSet.py
+++ b/avaframe/com1DFA/deriveParameterSet.py
@@ -390,24 +390,24 @@ def setThicknessValueFromVariation(key, cfg, simType, row):
                 for count, id in enumerate(idList):
                     thNameId = thType + id
 
-                if varType == 'Percent':
-                    cfg['GENERAL'][thNameId] = str(float(thicknessList[count]) * variationFactor)
-                    # set percentVaration parameter to actual variation in percent$steps
-                    if variationFactor == 1.:
-                        # if variation is 0% set percentVaration = ''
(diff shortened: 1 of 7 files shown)


matthiasto (approved):

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -354,45 +368,58 @@ def setThicknessValueFromVariation(key, cfg, simType, row):
...
+                        elif  variationFactor > 1:
+                            variationIni = '+' + str((variationFactor - 1.) * 100) + '$1'
+                    elif varType == 'Range':
+                        cfg['GENERAL'][thNameId] = str(float(thicknessList[count]) + variationFactor)
+                        # set percentVaration parameter to actual variation in percent$steps
+                        if variationFactor == 0.:
+                            # if variation is 0 set RangeVaration = ''
+                            variationIni = ''
+                        else:
+                            variationIni = str(variationFactor) + '$1'
```
fso42, line 410: I think this is now one step too much on the nesting and should be rewritten with subfunctions. Maybe seperate the vartypes?

fso42 (changes requested):
Please add info about this to the documentation right away.

avaframe/com1DFA/deriveParameterSet.py
awirb: i made subfunction for the creation of varIni

fso42 (approved):

fso42 (approved):

PR #659, opened 2022-02-02
fix #653 fix #654 AIMEC plot and runout

### LINKED ISSUE #653: aimec plot
add runout mean, max and min runout values in text in plot: slComparisonStat

### LINKED ISSUE #654: aimec runout point
assign 'real' x and y coordinates to runout point not the ones projected onto the path - so the x and y location can give an insight on how far off they are from each other

matthiasto:
Should work when the filtering parameter is a string
avaKot1_ppr_thresholdValue_1_slComparisonStat
avaKot1_relMaxppr_thresholdValue1
But also if it is a float
avaKot1_ppr_thresholdValue_1_slComparisonStat
avaKot1_relMaxppr_thresholdValue1


awirb (changes requested):
max ppr plot - maybe move colorbar to the left side of the plot so that there is no overlay between legend and colorbar

fso42 (approved):

PR #662, opened 2022-02-09
update probRun docu

awirb:


docs/moduleAna4Stats.rst
```diff
@@ -10,7 +10,7 @@ probAna
 probAna is used to derive simple probability maps for a set of simulations for one avalanche track.
 These maps show for each point in space the probability for a chosen parameter to exceed a given threshold.
 For example, it is possible to compute the probability map of an avalanche to exceed a
-peak pressure of 1kPa.
+peak pressure of 1kPa, but is also possible to chose other paramters and threshold values.
```
fso42, line 13:
```suggestion
peak pressure of 1kPa, but is also possible to chose other parameters and threshold values.
```

docs/moduleAna4Stats.rst
```diff
@@ -40,11 +40,21 @@ In order to run this example:
...
 
 
-.. figure:: _static/avaHockeyChannel_probMap.png
+.. figure:: _static/avaHockeyChannel_probMap_lim1.0.png
     :width: 90%
 
       Probability map example.
 
+Another example is given in :py:mod:`runScripts.runProbAnaCom1DFA.py`, but here only
+one parameter is varied at a time.
```
fso42, line 49:
```suggestion
An example utilising com1DFA is given in :py:mod:`runScripts.runProbAnaCom1DFA.py`, with only one parameter being varied at a time.
```

docs/moduleAna4Stats.rst
```diff
@@ -40,11 +40,21 @@ In order to run this example:
...
+Avalanche simulations are performed with the settings defined in the configuration file of
+:py:mod:`com1DFA` and in addition a parameter variation is performed according to the parameters
+set in ``ana4Stats/probAnaCfg.ini`` in the section PROBRUN. All the parameters set in PROBRUN are
+varied on at a time, i.e. simulations are performed for the standard settings of all parameters,
+except the one parameter to be varied, subsequently the other variations are performed.
+In the beginning of the script, the filtering criteria for generating probability maps can be set.
```
fso42, line 55:
```suggestion
In the beginning of the scrip filtering criteria for the probability maps can be set.
```

fso42 (changes requested):

docs/moduleAna4Stats.rst
awirb: In both examples com1DFA is used to create the simulations and then create the probability maps using these simulations, the point I wanted to make here is that with this script - it is possible to vary one parameter at a time, and then set probability configurations for creation of the maps, several possible- but it is a good point that for both example scripts com1DFA is used

fso42 (approved):

PR #667, opened 2022-02-17
fix issues, Particles saving

matthiasto:
small issue in aimec
also in the ini file, we could not convert negative values to numeric
and now saving particles for first and last  times steps

### DIFF
--- a/avaframe/com1DFA/com1DFA.py
+++ b/avaframe/com1DFA/com1DFA.py
@@ -910,7 +910,7 @@ def initializeParticles(cfg, releaseLine, dem, inputSimLines='', logName='', rel
         particles, hPartArray = particleTools.initialiseParticlesFromFile(cfg, avaDir, releaseLine['file'].stem)
     else:
         # initialize random generator
-        rng = np.random.default_rng(int(float(cfg['seed'])))
+        rng = np.random.default_rng(getint(cfg['seed']))
 
         nPart = 0
         xPartArray = np.empty(0)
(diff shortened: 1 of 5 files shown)


avaframe/com1DFA/com1DFA.py
```diff
@@ -909,7 +910,7 @@ def initializeParticles(cfg, releaseLine, dem, inputSimLines='', logName='', rel
         particles, hPartArray = particleTools.initialiseParticlesFromFile(cfg, avaDir, releaseLine['file'].stem)
     else:
         # initialize random generator
-        rng = np.random.default_rng(int(cfg['seed']))
+        rng = np.random.default_rng(int(float(cfg['seed'])))
```
fso42, line 913: What is the reason for the float conversion? Do we actually want non-integer seeds?

awirb (approved):

avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -105,11 +105,13 @@ entTh =
 dt = 0.1
 # End time [s]
 tEnd = 400
+# only one of the two following options can be activated
 # to use a variable time step (defined by the CFL number)
 cflTimeStepping = False
 # to use a variable time step (time step depends on kernel radius)
 sphKernelRadiusTimeStepping = False
```
fso42, line 112: This looks like an easy way to accidentally mix things up. I suggest switching to one variable with options, i.e:

timeStepping = cflTime
or 
timeStepping = sphKernelRadius

fso42 (changes requested):

avaframe/com1DFA/com1DFA.py
matthiasto: no the issue was if you ask for seeds in the ini file with 0:9:10 so you expect ['0', '1',' 2'...'9'] but you get ['0.0', '1.0'...'9.0']  and the conversion to int did not work. I hat to go first to float and then int, not sure why...

avaframe/com1DFA/com1DFACfg.ini
awirb: I think in this case it needs adding an option for using just dt as a fixed time step now which is our default currently

Issue #670, opened 2022-02-22
add simulation DEM if remeshed to different cellSize


awirb:
* create folder Inputs/DEM
* add remeshed DEM as ascii file there with name: remeshedDEM%.2f % cellSize 
* create flag in ini file to search for available DEMs 
* create function to first check cellSize of standard DEM in Inputs 
-> if not right cellSize go to Inputs/DEM check one after the other (also if not correct name)

awirb:
solved with PR #677 

* create folder Inputs/DEMremeshed 
* add remeshed DEM as ascii file there with name: DEMINPUTSNAME_remeshedDEM%.2f.asc % cellSize
* create flag in ini file to clean Inputs/DEMremeshed - default is true
* this folder cleaned only ONCE per call to  com1DFAMain 
* create function to first check cellSize of standard DEM in Inputs
    -> if not right cellSize go to Inputs/DEMremeshed check one after the other if matching cellSize and DEMINPUTSNAME in name - use this one, if not found remesh and save

Issue #658, opened 2022-02-01
Enable to chose dem asc file for com1DFA


matthiasto:
When doing simulations with varying cell size, the remeshing step can be time consuming. An option to precompute the dem on different cell sizes and then during com1DFA run, choosing the good one could save some time.

fso42:
So add a write out to geoTrans.remesh and replace the .asc in the project/input would be a viable solution?

PR #675, opened 2022-02-28
check for remeshed DEM, save remeshed DEM

awirb:
* if desired meshCellSize in configuration doesn't match approx. cellSize of DEM in Inputs/ - go to DEMremeshed and load one .asc file after the other if a DEM found with good cellSize -> stop and use this one
* if none found -> remesh 
* if remeshing is performed -> save new DEM to Inputs/DEMremeshed as remeshedDEM%.2f.asc % meshCellSize
* close issue #670


avaframe/in3Utils/geoTrans.py
```diff
@@ -288,35 +291,85 @@ def remeshDEM(cfg, dem):
...
+            flag if dem is found
+    """
+
+    # load input parameters
+    cellSize = cfg.getfloat('meshCellSize')
+    cellSizeThreshold = cfg.getfloat('meshCellSizeThreshold')
+
+    # path to remeshed DEM folder
+    pathToDem = pathlib.Path(cfg['avalancheDir'], 'Inputs', 'DEMremeshed')
+    fU.makeADir(pathToDem)
```
matthiasto, line 360: Why make the folder? You only need to read it no?

avaframe/in3Utils/geoTrans.py
```diff
@@ -288,35 +291,85 @@ def remeshDEM(cfg, dem):
...
+    demFiles = list(pathToDem.glob('*.asc'))
+    DEMFound = False
+    dem = {}
+    for demF in demFiles:
+        demDict = IOf.readRaster(demF)
+        if abs(cellSize - demDict['header']['cellsize']) < cellSizeThreshold:
+            log.info('Remeshed DEM found: %s cellSize: %.5f' % (demF.name, demDict['header']['cellsize']))
+            DEMFound = True
+            dem = demDict
+            continue
```
matthiasto, line 372: suggestion. Maybe we could make kind of a csv file with the info about all remeshed dem. Then you could look at this dataframe to choose the good dem?

avaframe/in3Utils/geoTrans.py
```diff
@@ -288,35 +291,85 @@ def remeshDEM(cfg, dem):
...
-        dem['header'] = headerRemeshed
-        xNewGrid, yNewGrid = np.meshgrid(xNew, yNew)
-        zNew = sp.interpolate.griddata((xGrid, yGrid), z, (xNewGrid, yNewGrid), method='cubic',
-                                       fill_value=headerDEM['noDataValue'])
-        log.info('Remeshed data extent difference x: %f and y %f' % (diffExtentX, diffExtentY))
-        dem['rasterData'] = zNew
-
-    return dem
+        # first check if remeshed DEM is available
+        remeshedDem, DEMFound = searchRemeshedDEM(cfg)
```
matthiasto, line 295: maybe add an option to force the remeshing even if the re meshed dem already exists? And make it the default one.

matthiasto (changes requested):

avaframe/in3Utils/geoTrans.py
awirb: how should we do this? every time a remeshedDEM is stored we open the csv and append the info? but then it is not allowed to store a DEM from outside there, and if we read the DEM first to create the csv. it takes more time

fso42: I´m not a fan of creating a csv! We are not talking about a common occurance with tons of files. Reading a dem and getting the header is not huge expense...

avaframe/in3Utils/geoTrans.py
awirb: this directory only exists if before a remeshed DEM has been saved, hence when searching - I changed it to check if dir exists

avaframe/in3Utils/geoTrans.py
awirb: I added a flag forceRemesh in com1DFACfg.ini - if True force remeshing - now default

avaframe/com1DFA/com1DFA.py
```diff
@@ -267,8 +267,7 @@ def com1DFACore(cfg, avaDir, cuSimName, inputSimFiles, outDir):
     # +++++++++PERFORM SIMULAITON++++++++++++++++++++++
     # for timing the sims
     startTime = time.time()
-    particles, fields, dem, reportAreaInfo = initializeSimulation(cfg, demOri, inputSimLines, cuSimName)
-
+    particles, fields, dem, reportAreaInfo, demOri = initializeSimulation(cfg, demOri, inputSimLines, cuSimName)
```
fso42, line 270: I think the logic for demOri vs dem needs to be tackled:
- If there is demOri as input, but an additional dem as output, I think the basic (and valid) assumption is, that demOri remains unchanged and `dem` contains the topography to use -> then there's no need to return demOri
- If demOri is somehow changed, why then the need for an additional `dem` variable?

I suggest to make it a clear distinction: demOri -> the originial, untouched DEM, dem -> the applied, and maybe modified, dem

avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -198,11 +198,13 @@ cMaxnMass = 5
...
 # minimum flow depth [m]
 hmin = 0.05
-# remesh the input DEM
+# remesh the input DEM or look for remeshed DEM 
 # expected mesh size [m]
 meshCellSize = 5
 # threshold under which no remeshing is done
 meshCellSizeThreshold = 0.001
+# force remesh - if True remeshing performed also if remeshed DEMs in Inputs/DEMremeshed available
+forceRemesh = True
```
fso42, line 207: why is a forceRemesh necessary? Either the mesh is different from the specified resolution -> remesh, or it is the same -> do not remesh...

avaframe/in3Utils/geoTrans.py
```diff
@@ -286,8 +289,16 @@ def remeshDEM(cfg, dem):
...
-    # remesh if input DEM size does not correspond to the computational cellSize
-    if np.abs(cszNew - cszDEM) > cszThreshold:
+    remesh = True
+    if cfg.getboolean('forceRemesh') is False:
+        # first check if remeshed DEM is available
+        remeshedDEM, DEMFound = searchRemeshedDEM(cfg, dem)
+        if DEMFound:
+            remesh = False
+    else:
+        log.info('Forcing remeshing')
```
fso42, line 299: see comment about forceRemesh

avaframe/in3Utils/geoTrans.py
```diff
@@ -286,8 +289,16 @@ def remeshDEM(cfg, dem):
...
+
+    if remesh:
```
fso42, line 301: drop this  if and add a return above

avaframe/in3Utils/geoTrans.py
```diff
@@ -309,14 +320,68 @@ def remeshDEM(cfg, dem):
...
+    cellSize = cfg.getfloat('meshCellSize')
+    cellSizeThreshold = cfg.getfloat('meshCellSizeThreshold')
+
+    # path to remeshed DEM folder
+    pathToDem = pathlib.Path(cfg['avalancheDir'], 'Inputs', 'DEMremeshed')
+    DEMFound = False
+    remeshedDEM = {}
+    if pathToDem.is_dir():
+        # look for dems and check if cellSize within tolerance and origin matches
+        demFiles = list(pathToDem.glob('*.asc'))
```
fso42, line 369: split this here and move to for to the same indent as the next for. Do a else return/etc on the if pathToDem

avaframe/in3Utils/geoTrans.py
```diff
@@ -309,14 +320,68 @@ def remeshDEM(cfg, dem):
...
+        for demF in demFiles:
+            headerDEM = IOf.readASCheader(demF)
+            if abs(cellSize - headerDEM['cellsize']) < cellSizeThreshold:
+                if ('%.2f' % headerDEM['xllcenter']) == ('%.2f' % dem['header']['xllcenter']) and ('%.2f' % headerDEM['yllcenter']) == ('%.2f' % dem['header']['yllcenter']):
+                    remeshedDEM = IOf.readRaster(demF)
+                    log.info('Remeshed DEM found: %s cellSize: %.5f' % (demF.name, remeshedDEM['header']['cellsize']))
+                    DEMFound = True
+                    continue
+                else:
+                    log.warning('Remeshed dem found with cellSize %.2f but origin not matching' % cellSize)
```
fso42, line 379: Is checking cellsize and origin really enough? Should we check the name as well? Is the origin even needed?

fso42 (changes requested):

fso42:
Maybe consider #658  as well?

awirb:
see new PR #677

PR #692, opened 2022-03-17
Add ana5Utils

awirb:
distanceTimeAnalysis

* create distance-time diagrams of ava simulations from a reference point showing the avalanche front and the average values of a chosen result parameter (e.g. flow depth, flow velocity) 
* option 1: tt-diagram reference point is beta point
* option 2: simulated range-time diagram reference point radar location, masked to radar field of view
* perform coordinate transformation of simulation results for I) along avalanche path (based on aimec functions) or use ii) radar’s coordinate system to identify avalanche front and average values of flow parameter for all range gates
Code is based on the code for creating simulated range – time diagrams with respect to the radar’s field of view from Anselm Köhler @snowtechblog  and Michael Winkler and additional functionalities (e.g. tt-diagram) have been added. 
The ideas of presenting avalanche simulation results in a radar’s coordinate system are based on previous work presented in Fischer et al. (2014) and Rauter and Köhler (2019), see documentation ana5Utils chapter for full references.


avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+        --------
+        maxVel: float
+            max value of approach velocity
+    """
+
+    # load lists
+    timeList = mtiInfo['timeList']
+    rangeList = mtiInfo['rangeList']
+
+    # as these might be not in ascending order timestep wise - order first
```
matthiasto, line 773: why not?

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+    # nanmin required as if FV is zero range is nan
+    rangeListSorted = np.asarray(rangeList)[np.argsort(timeList)] + abs(np.nanmin(np.asarray(rangeList)))
+    approachVelocity = []
+    maxVel = 0.0
+    rangeVel = 0.0
+    # compute deltaDistance/deltaTime to get approach velocity
+    # use interval of 2*dtSave for little smmoothing
+    for index, range in enumerate(rangeListSorted[1:-1]):
+        vel = (rangeListSorted[index+1] - rangeListSorted[index-1]) / (timeListSorted[index+1] - timeListSorted[index-1])
+        approachVelocity.append(np.abs(vel))
```
matthiasto, line 786: why not also return the approachVelocity? if not returning it why do you store it?

docs/references_all.bib
```diff
@@ -20194,3 +20194,15 @@ @ARTICLE{ZwKlSa2003
   pages = {161--194},
   file = {ZwKlSa2003_SAMOS_Springer.pdf:SAMOS_LIT/ZwKlSa2003_SAMOS_Springer.pdf:PDF}
 }
+
+
+@article{RaKo2019,
```
matthiasto, line 20199: isnt it 2020?

avaframe/out3Plot/outDistanceTimeAnalysis.py
```diff
@@ -0,0 +1,398 @@
...
+    rangeTimeResType = cfgRangeTime['rangeTimeResType']
+    maxVel, rangeVel, timeVel = dtAna.approachVelocity(mtiInfo)
+
+    # in case time steps are not ordered - the colormesh x and y need to be ordered
+    timeIndex = np.argsort(np.array(timeList))
+    timeListNew = np.array(timeList)[timeIndex]
+    mti = mti[:, timeIndex]
+
+    # create plot
+    fig = plt.figure()
```
matthiasto, line 48: maybe add our pU.figWidth...

avaframe/tests/test_distanceTimeAnalysis.py
```diff
@@ -0,0 +1,168 @@
...
+    Y = 10.
+    Z = np.sqrt(10.**2 + 10.**2)
+
+    # call function to be tested
+    r, phi, theta = dtAna.cartToSpherical(X, Y, Z)
+
+    assert r == np.sqrt(10.**2 + 10.**2 + Z**2)
+    assert phi == 45.
+    assert theta == 45.
+
```
matthiasto, line 84: I know it is annoying but try with something else than 45 degrees to be sure that we do not have issues with the cos and sin (cos(45)=sin(45) so you could have an issue without being aware of it...)

avaframe/tests/test_distanceTimeAnalysis.py
```diff
@@ -0,0 +1,168 @@
...
+    assert demOriginal['header']['nrows'] == 11
+    assert demOriginal['header']['cellsize'] == 2.
+
+
+def test_rotate():
+    """ test rotate a vector by an angle """
+
+    # setup required data
+    locationPoints = [[0., 0.], [0., 10.]]
+    theta = 45.
```
matthiasto, line 54: same here, 45 is maybe not the best idea

matthiasto (changes requested):

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
```diff
@@ -0,0 +1,48 @@
...
+
+
+[GENERAL]
+# result type used for creating range time diagram e.g. FD, FV, ....
+rangeTimeResType = FD
+# time steps used for creating diagram
+tSteps = 0:1
+# threshold - below this result data is ignored - used to create mask
+# average value created of where there is avalanche flow (threshold > 0.0) or including all cells in
+# range gate (also the ones where there is no snow flowing - threshold = 0.0)
```
matthiasto, line 13: should we do it the other way around so that we keep field>threshold. Then we do not have to do this trick of putting 0.000001

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+    # fetch radar field of view coordinates
+    radarFov = getRadarLocation(cfgRangeTime)
+
+    # create masked array of radar range with DEM extent using radar field of view and range gates
+    rangeMasked, rangeGates = radarMask(dem, radarFov, cfgRangeTime['GENERAL'].getfloat('aperture'),
+        cfgRangeTime)
+
+    # setup time and distance to radar from avalanche front lists
+    timeList = []
+    rList = []
```
matthiasto, line 107: This is unused

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,99 @@
...
+
+With the functions gathered in this module, flow parameters of avalanche simulation results can be
+visualized in a distance versus time diagram, the so called **thalweg-time diagram**.
+The **tt-diagram** provides a new way to identify main features of the temporal evolution of
+a flow parameter along the :term:`avalanche path`.
+This is based on the ideas presented in :cite:`FiFrGaSo2013` and :cite:`RaKo2019`, where
+avalanche simulation results have been transformed into the radar coordinate system to facilitate
+direct comparison.
+In addition to the **tt-diagram**, :py:mod:`ana5RangeTime` also offers the possibility to
+produce simulated **range-time diagrams** of the flow parameters with respect to a radar's field
```
jtfisch, line 18:
```suggestion
produce simulated **range-time diagrams** of the flow variables with respect to a radar's field
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,99 @@
...
+~~~~~~~~~
+
+Thalweg-time diagram
+~~~~~~~~~~~~~~~~~~~~~~
+
+First, the flow parameter result field is transformed into a path-following coordinate system.
+For this step, functions from :py:mod:`ana3AIMEC` are used.
+The distance to the *start of runout area point* is determined and the mean values of the cross
+profiles along the avalanche path are computed.
+In the **tt-diagram**, the avalanche front position is indicated with black dots and the colored
```
jtfisch, line 77: how is front defined?
front, which is defined via a threshold value ... ?

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,99 @@
...
+field indicates the mean values of the flow parameter at cross profiles along the avalanche path
+for each time step.
```
jtfisch, line 79: for which rasters / flow parameters can a TT diagram be created? all - whats the difference ? is the front always defined with respect to the analyzed aparameter?

jtfisch: whort info on the depicted velocities: in which direction are they measured (approach velocities in projected avalanche path), particularly with respect to the flow velocities (surface parallel?). What does the location of the max front velocity indicate?

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,99 @@
...
+
+
+
+Simulated Range-Time diagram
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The radar's field of view is determined using its location, a point in the direction of the field of
+view and the aperture angle. The flow parameter result field is transformed in the radar coordinate
```
jtfisch, line 87:
```suggestion
view and the aperture angle. The flow variable result field is transformed in the radar coordinate
```

jtfisch (changes requested):

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,99 @@
...
+distanceTimeAnalysis: Visualizing the temporal evolution of flow parameters
+------------------------------------------------------------------------------
+
+With the functions gathered in this module, flow parameters of avalanche simulation results can be
+visualized in a distance versus time diagram, the so called **thalweg-time diagram**.
+The **tt-diagram** provides a new way to identify main features of the temporal evolution of
+a flow parameter along the :term:`avalanche path`.
+This is based on the ideas presented in :cite:`FiFrGaSo2013` and :cite:`RaKo2019`, where
+avalanche simulation results have been transformed into the radar coordinate system to facilitate
+direct comparison.
```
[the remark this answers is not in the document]
awirb: added this

docs/moduleAna5Utils.rst
awirb: I updated the paragraph

avaframe/tests/test_distanceTimeAnalysis.py
awirb: also use different angle now for testing

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
awirb: any suggestions?

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+    # mask this with full mask (result below threshold and outside field of view)
+    radarRange = np.ma.array(r, mask=maskFull)
+
+    # line of sight min of masked array
+    losRange = radarRange.min()
+
+    return losRange
+
+
+def fetchFlowFields(flowFieldsDir, suffix=''):
```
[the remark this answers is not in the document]
fso42: If not, then move it to a general module

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+    threshold = cfgRangeTime['GENERAL'].getfloat('thresholdResult')
+    rangeGates = mtiInfo['rangeGates']
+    Rarray = mtiInfo['Rarray']
+    bmaskRadar = mtiInfo['bmaskRadar']
+    rangeMasked = mtiInfo['rangeMasked']
+    mti = mtiInfo['mti']
+    mtiNew = np.zeros((len(rangeGates), 1))
+
+    # get mask of results below threshold
+    maskAva = avalancheMask(flowF, threshold)
```
fso42, line 417: see my note on the avalancheMask function

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+    rasterTransfo = mtiInfo['rasterTransfo']
+
+    # add header to result field - required for coordinate transformation
+    fieldFull = {'header': demHeader, 'rasterData': flowF}
+
+    # transorm result field into avalanche path following coordinate system
+    slRaster = aT.transform(fieldFull, rasterTransfo, cfgRangeTime['GENERAL']['interpMethod'])
+
+    # fetch raster area and compute mean, max values for each cross-profile
+    # TODO: average over cells – weighted will cell area but only projected area (aimec function)
```
fso42, line 594: Address todo right away?

avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -304,6 +304,12 @@ particleProperties =
 writePartToCSV = False
 # particle properties to be saved to csv (options: ux, uy, uz, velocityMagnitude,..)
 particleProperties = velocityMagnitude|m
+# create range-time diagram of simulation results -
+# additional configurations in ana5Utils - distanceTimeAnalysisCfg
+createRangeTimeDiagram = False
+# if TTdiagram=True - create thalweg-time diagram,
+# if False create range-time diagram with radar field of view
+TTdiagram = True
```
fso42, line 312: I'm really against making/controlling this from a INI file! This is visualisation and should be done AFTER the model. Especially since we want to have it applicable to more models.
If someone wants to plott the TTdiag, they run the model and THEN do the analysis /plotting. This Section of the config should contain stuff that might be needed FOR visualisation, but it should not trigger the analysis / visualisation

avaframe/out3Plot/outDistanceTimeAnalysis.py
```diff
@@ -0,0 +1,398 @@
...
+    cName = 'range'
+    pU.addColorBar(pc, ax1, None, unit, title=cName)
+    pU.putAvaNameOnPlot(ax1, cfgRangeTime['avalancheDir'])
+    pU.putInfoBox(ax1, '- range gates [%d]' % gateContours, location='lowerRight', color='brown')
+
+    # create plot of DEM and radar field of view
+    fig2 = plt.figure(figsize=(pU.figW, pU.figH))
+    # add second panel with DEM and radar location
+    ax2 = plt.subplot(111)
+    #pc0 = plt.pcolormesh(X, Y, dem['rasterData'], cmap=cm.grayC)
```
fso42, line 139: remove

avaframe/runScripts/runRangeTimeDiagram.py
```diff
@@ -0,0 +1,88 @@
+"""
+    Run script for module com2AB
```
fso42, line 2: Change

avaframe/ana3AIMEC/aimecTools.py
```diff
@@ -565,8 +565,8 @@ def transform(fname, rasterTransfo, interpMethod):
 
     Parameters
     ----------
-    fname: str
-        path to rasterfile to transform
+    fname: str, dict
+        path to rasterfile to transform- or already read to dict with data as np nd array
```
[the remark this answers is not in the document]
awirb: yes I often use the shortcut how we import the module, I changed it to numpy

avaframe/ana3AIMEC/aimecTools.py
```diff
@@ -1059,3 +1064,199 @@ def getMaxMeanValues(rasterdataA, rasterArea):
...
+        Paramters
+        ----------
+        rasterTransfo: dict
+            domain transformation info
+
+        Returns
+        --------
+        rasterTransfo: dict
+            updated dictionary with new coordinate system
+    """
```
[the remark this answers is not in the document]
awirb: done

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+
+def setDemOrigin(demSims):
+    """ reset original origin to simulation DEM - required if for ava sim is set to something different
+
+        Parameters
+        ------------
+        demSims: dict
+            dictionary with header and data of DEM used to run ava simulations
+            must contain a key called originalHeader - where the original origin is given
+
```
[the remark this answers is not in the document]
awirb: added an error with: DEM dictionary does not have originalHeader key - required to get original origin

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+    """ function to create masked array of radar range using dem and radar's field of view
+
+        Parameters
+        -----------
+        dem: dict
+            DEM dictionary with header info and rasterData
+        radarFov: list
+            radars field of view min and max points
+        aperture: float
+            opening angle?
```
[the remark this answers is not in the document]
awirb: done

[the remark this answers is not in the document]
awirb: changed to aperture angle here

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+        -------------
+        flowF: numpy nd array
+            flow parameter result field
+        rangeMasked: masked numpy array
+            masked array of radar range
+        threshold: float
+            used to create mask for getting line of sight range,
+            data below threshold not taken into account
+
+        outfn:
```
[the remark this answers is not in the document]
awirb: removed

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+            updated mtiInfo dict with info on mti values
+    """
+
+    # load parameters from configuration file and mtiInfo
+    rgWidth = cfgRangeTime['GENERAL'].getfloat('rgWidth')
+    # if threshold = 0.0 mean over all cells - if threshold > 0.0 mean only over all cells where there
+    # is avalanche flow
+    threshold = cfgRangeTime['GENERAL'].getfloat('thresholdResult')
+    rangeGates = mtiInfo['rangeGates']
+    Rarray = mtiInfo['Rarray']
```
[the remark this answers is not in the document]
awirb: changed

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+        dem: dict
+            dictionary with DEM header and data
+        simHash: str
+            unique simulation ID
+
+        Returns
+        --------
+        mtiInfo: dict
+            info on time steps (timeList) and distance (rangeList) of avalanche front
+            averaged values of result parameter (mti) for each range gate (rangeGates) for colorcodig
```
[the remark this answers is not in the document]
awirb: colorcoding

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
```diff
@@ -0,0 +1,48 @@
...
+
+#+++++range time diagram with radar field of view +++++
+# width of range gates in meters
+rgWidth = 5.
+# every gateContours gate range plotted as contour line in radar field of view plot
+gateContours = 20
+# aperture angle in degrees
+aperture = 15
+
+# radar location field of view points min (x0, y0) and point giving direction (x1, y1) format: x0|x1|y0|y1
```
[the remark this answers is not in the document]
awirb: added this

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,102 @@
...
+ana5Utils
+#####################################################################
+
+
+distanceTimeAnalysis: Visualizing the temporal evolution of flow variables
+------------------------------------------------------------------------------
+
+With the functions gathered in this module, flow variables of avalanche simulation results can be
+visualized in a distance versus time diagram, the so called **thalweg-time diagram**.
+The **tt-diagram** provides a new way to identify main features of the temporal evolution of
```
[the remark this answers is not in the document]
awirb: removed new

avaframe/ana5Utils/distanceTimeAnalysis.py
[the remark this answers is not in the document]
awirb: readFields really reads the data of the ascii files, here I just want to have the list of file paths - I moved it to fileHandlerUtils

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
matthiasto: instead of masking cell with value <= threshold keep cells with value > threshold

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: made an issue as interesting for aimec too: https://github.com/avaframe/AvaFrame/issues/695

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+    rangeGates= np.arange(minR+rgWidth/2, maxR+rgWidth/2, rgWidth)
+
+    # create plot of range distance already masked with radar field of view
+    dtAnaPlots.radarFieldOfViewPlot(radarFov, aperture, radarRange, X, Y, cfgRangeTime['GENERAL'],
+        rangeGates, demOriginal)
+
+    return radarRange, rangeGates
+
+
+def avalancheMask(resData, threshold):
```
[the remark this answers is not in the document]
awirb: it is not where equals threshold but where smaller than threshold so I changed it to use: https://numpy.org/doc/stable/reference/generated/numpy.ma.masked_where.html#numpy.ma.masked_where

avaframe/ana3AIMEC/aimecTools.py
```diff
@@ -1059,3 +1064,199 @@ def getMaxMeanValues(rasterdataA, rasterArea):
...
+        --------
+        avaPath: dict
+            info on aimec path coordinates, ...
+        splitPoint: dict
+            info on split point coordinates
+
+    """
+
+    # fetch input parameters
+    ProfileLayer = pathDict['profileLayer']
```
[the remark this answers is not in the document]
awirb: changed this

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+
+    # add all info to mti dict
+    mtiInfo = {'mti': mti, 'rangeGates': rangeGates, 'Rarray': Rarray, 'bmaskRadar': bmaskRadar,
+        'rangeList': [], 'timeList': [], 'rangeMasked': rangeMasked,
+        'demOriginal': dem, 'type': 'rangeTime', 'referencePointName': 'radar'}
+
+    return mtiInfo
+
+
+def rotate(locationPoints, theta, deg=True):
```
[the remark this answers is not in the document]
awirb: moved it to geoTrans and made numpy array

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: removed it

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+    rotatedLine = [[locationPoints[0][0], float(locationPoints[0][0]+vectorRot[0])], # x
+           [locationPoints[1][0], float(locationPoints[1][0]+vectorRot[1])] #y
+          ]
+
+    # check if rotation worked correctly
+    lenVector = np.sqrt(vector[0]**2 + vector[1]**2)
+    lenVectorRot = np.sqrt(vectorRot[0]**2 + vectorRot[1]**2)
+    vecDiff = abs(lenVector - lenVectorRot)
+    if vecDiff > 1.e-7:
+        log.warning('rotated vectors are not of same length %.2f vs %.2f' % (lenVector, lenVectorRot))
```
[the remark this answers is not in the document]
awirb: not accessible in pytest

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+
+    """
+
+    # fetch coordinate info
+    locationInfo = cfg['GENERAL']['radarLocation'].split('|')
+
+    # set radar location and point in direction of field of view as list
+    radarFov = [[float(locationInfo[0]), float(locationInfo[1])], [float(locationInfo[2]),
+        float(locationInfo[3])]]
+
```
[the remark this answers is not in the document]
awirb: now throw an error if outside

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+    # demOriginal['header']['noDataValue'] = -9999
+    # radarRangeNew = np.where(maskPhi == False , radarRange, -9999)
+    # outFile = pathlib.Path(cfgRangeTime['GENERAL']['avalancheDir'], 'radarRange.asc')
+    # IOf.writeResultToAsc(demOriginal['header'], radarRangeNew, outFile, flip=True)
+
+    # create range gates
+    minR = int(np.floor(np.nanmin(radarRange)))
+    maxR = int(np.ceil(np.nanmax(radarRange)))
+    rgWidth = cfgRangeTime['GENERAL'].getfloat('rgWidth')
+    rangeGates= np.arange(minR+rgWidth/2, maxR+rgWidth/2, rgWidth)
```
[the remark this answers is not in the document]
awirb: I think this is not a problem as it just creates an array where the last one is for sure inside

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+    for indexRI, rI in enumerate(rangeGates):
+        if rI < minRAva or rI > maxRAva:
+            # do not expect to see data for this range
+            continue
+        # create mask for range slice within range gate
+        bmaskRange = ~np.logical_and(Rarray > rI-rgWidth/2, Rarray < rI+rgWidth/2)
+        bmaskAvaRadarRangeslice = ~np.logical_and(~bmaskRange, ~bmaskAvaRadar)
+        if np.any(~bmaskAvaRadarRangeslice):
+            # only update if range_slice mask is not empty
+            mtiValue = np.mean(np.ma.array(np.ma.getdata(maskAva), mask=bmaskAvaRadarRangeslice))
```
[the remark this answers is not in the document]
awirb: changed this to only loop over: smallRangeGatesIndex = np.where((rangeGates > minRAva) & (rangeGates < maxRAva))[0]

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,792 @@
...
+        #TODO why int?
+        minRAva = int(np.floor(rmaskAva_radar.min()))
+        maxRAva = int(np.ceil(rmaskAva_radar.max()))
+    else:
+        minRAva = np.amin(rangeGates)
+        maxRAva = np.amax(rangeGates)
+
+    # loop over each range gate to populate mti array
+    for indexRI, rI in enumerate(rangeGates):
+        if rI < minRAva or rI > maxRAva:
```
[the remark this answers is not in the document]
awirb: changed to loop over: smallRangeGatesIndex = np.where((rangeGates > minRAva) & (rangeGates < maxRAva))[0]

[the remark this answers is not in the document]
awirb: and with only looping if below maxRava shouldn't be a problem

avaframe/ana3AIMEC/aimecTools.py
```diff
@@ -1059,3 +1064,200 @@ def getMaxMeanValues(rasterdataA, rasterArea):
...
+    This function returns the information about the domain transformation
+    Data given on a regular grid is projected on a nonuniform grid following
+    a polyline to end up with "straightend raster"
+
+    Parameters
+    ----------
+    avaDir: pathlib path or str
+        path to avalanche directory
+    dem: dict
+        dictionary with header and raster data
```
matthiasto, line 1200: This is almost the same as makeDomainTrans.
Could we try to only have one function. I think it should be possible. 
If not (but I would prefer to avoid it) can you use the new sub functions (setAvaPath...) in makeDomainTrans

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,653 @@
...
+import avaframe.out3Plot.plotUtils as pU
+import avaframe.com1DFA.com1DFA as com1DFA
+import avaframe.in3Utils.fileHandlerUtils as fU
+
+# create local logger
+# change log level in calling module to DEBUG to see log messages
+log = logging.getLogger(__name__)
+
+
+def getRadarLocation(cfg):
```
matthiasto, line 31: is def splitIniValueToArraySteps(cfgValues, returnList=False) not doing exactly this?

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,653 @@
...
+
+    # get mask of results below threshold
+    maskAva = np.ma.masked_where(flowF < threshold, flowF)
+    bmaskAva = np.ma.getmask(maskAva)
+    # and join to a mask with "visible part of avalanche" - mask where not field of view and not
+    # ava result above threshold
+    bmaskAvaRadar = ~np.logical_and(~bmaskRadar, ~bmaskAva)
+    # mask range with this mask to obtain 'visible part of avalanche' ranges
+    rmaskAva_radar= np.ma.array(np.ma.getdata(rangeMasked), mask=bmaskAvaRadar) # masked ranges
+
```
matthiasto, line 286: this is exactly the same as minRangeSimulation(). Could we keep the rmaskAva_radar saved so that we can reuse it?

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,653 @@
...
+        Returns
+        --------
+        losRange: float
+            minimum distance in line of sight to avalanche front
+            defined by the flowF
+
+    """
+
+    # fetch mask where result field smaller than threshold
+    maskAva = np.ma.masked_where(flowF < threshold, flowF)
```
matthiasto, line 228: I guess if you use 
maskAva = np.ma.masked_where(flowF <= threshold, flowF)
and have a threshold = 0
You do not have to trick with threshold = 0.00001
This would apply everywhere

avaframe/tests/test_distanceTimeAnalysis.py
```diff
@@ -0,0 +1,103 @@
...
+
+def test_radarMask(tmp_path):
+    """ test creating radar mask array """
+
+    # setup required input
+    testAvaDir = pathlib.Path(tmp_path, 'testAva')
+    headerDEM = {'xllcenter': 0.0, 'yllcenter': 0.0, 'cellsize': 1., 'ncols': 11, 'nrows': 11}
+    demOriginal = {'header': headerDEM, 'rasterData': np.zeros((11, 11))}
+    radarFov = [[1., 10.0], [5., 5.]]
+    aperture = 45.
```
matthiasto, line 57: could we try to use other angles as 45?

avaframe/tests/test_geoTrans.py
```diff
@@ -460,3 +460,72 @@ def getIPZ(z0, xEnd, yEnd, dx):
...
+    Y = 10.
+    Z = np.sqrt(10.**2 + 10.**2)
+
+    # call function to be tested
+    r, phi, theta = geoTrans.cartToSpherical(X, Y, Z)
+
+    assert r == np.sqrt(10.**2 + 10.**2 + Z**2)
+    assert phi == 45.
+    assert theta == 45.
+
```
matthiasto, line 479: same here: could we try to use other angles as 45? 45 is a very bad choice in my opinion

docs/references_all.bib
```diff
@@ -20194,3 +20194,30 @@ @ARTICLE{ZwKlSa2003
...
+}
+
+
+
+@Article{RaKo2020,
+AUTHOR = {Rauter, Matthias and K{\"o}hler, Anselm},
+TITLE = {Constraints on Entrainment and Deposition Models in Avalanche Simulations from High-Resolution Radar Data},
+JOURNAL = {Geosciences},
+VOLUME = {10},
+YEAR = {2020},
```
matthiasto, line 20217: why twice the same?

matthiasto (changes requested):

avaframe/tests/test_geoTrans.py
awirb: there is a second test case just the lines below where it is not 45, just kept both

docs/references_all.bib
awirb: removed

avaframe/ana3AIMEC/aimecTools.py
awirb: made an issue here

avaframe/tests/test_distanceTimeAnalysis.py
awirb: changed the angle

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: no, this function makes a 1D vector but I need here a 2x2 array

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,102 @@
...
+**range-time** diagrams refers to average values of the chosen flow parameter
+(e.g. flow depth, flow velocity) at specified range gates. This colorcoding is not directly
+comparable to the intensity given in the range-time diagram from radar measurements.
+
+.. Note::
+  The **tt-diagram** and the **range-time diagram** can be produced *during run time* of
+  :py:mod:`com1DFA`, or as a postprocessing step. However, the second option requires saving all the
+  required time steps of the flow parameter fields, which is much more computationally expensive
+  compared to the first option.
+
```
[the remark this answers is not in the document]
snowtechblog: Even if it is only a post-processing step, you would need to alter the output to write each timestep. I disagree because it produces so much overload and is only feasible for a couple of simulations.

avaframe/ana5Utils/distanceTimeAnalysis.py
snowtechblog: I depends a bit on the stability of the result field. Are we sure, that outside an avalanche the result field is always numerically 0? Or do sometime very small numbers exist? We could also do "maskAva = np.ma.masked_where(flowF <= threshold + epsilon, flowF) with threshold = 0 and epsilon = sys.float_info.epsilon" to prevent problems with float comparison (https://stackoverflow.com/a/9528651)

docs/moduleAna5Utils.rst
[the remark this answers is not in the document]
fso42: That is true for the generation of the data, but not for the plotting itself. Similar to saving particles, but doing the postprocessing/visualization i.e. with paraview...
I.e preparation/aggregation of data is fine... @awirb will change it as discussed in meeting.

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,653 @@
...
+             2) thalweg-time diagram from beta point of view along avalanche path
+"""
+
+
+import numpy as np
+import matplotlib as mpl
+from matplotlib import pyplot as plt
+import logging
+import pathlib
+
```
snowtechblog, line 13:
```suggestion
import sys
```

avaframe/ana5Utils/distanceTimeAnalysis.py
snowtechblog, line 228:
```suggestion
    epsilon =  sys.float_info.epsilon
    maskAva = np.ma.masked_where(flowF <= threshold + epsilon, flowF)
```

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
```diff
@@ -0,0 +1,49 @@
...
+
+[GENERAL]
+# result type used for creating range time diagram e.g. FD, FV, ....
+rangeTimeResType = FD
+# time steps used for creating diagram
+tSteps = 0:1
+# threshold - below this result data is ignored - used to create mask
+# average value created of where there is avalanche flow (threshold > 0.0) or including all cells in
+# range gate (also the ones where there is no snow flowing - threshold = 0.0)
+thresholdResult = 0.0001
```
snowtechblog, line 14:
```suggestion
thresholdResult = 0.0
```

avaframe/ana5Utils/distanceTimeAnalysis.py
snowtechblog: Please make sure this way of thresholding is at the other places as well. Thanks

snowtechblog (changes requested):

avaframe/com1DFA/com1DFACfg.ini
awirb: changed this to just produce the mtiInfo and save this to pickle and then it is available

avaframe/ana5Utils/distanceTimeAnalysis.py
fso42: Question about the threshold applied:
1. Is this always related to the machine precision? 
2. Or is it related to also include the impreciseness that comes from the model itself? Like eg for certain questions/tasks one ignores eg. everything where release thickness is below 10cm or something because for the question at hand those values are irrelevant.?

We are talking about relative vs absolute error  (good comparison here: https://codingnest.com/the-little-things-comparing-floating-point-numbers/#fn13). And while @snowtechblog  proposed solution is right from a technical / purely mathematical standpoint (under the assumption that all our results are precise down to machine precision), in reality this is irrelevant. Another good resource is https://floating-point-gui.de/errors/comparison/ 
There is a reason why functions like https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html have both rtol and atol as parameters. 

Or to rephrase it: is it even remotely relevant to look at something to the e-16 for closeness when we know that we introduce, as @snowtechblog said, those 'very small numbers' because of some numerical issue? like for example the 'unstoppable particles' due to friction (among many others)?

So what I propose is to use a tolerance threshold value that makes sense in regards to the data we are looking at. And maybe to make this more obvious the first proposal by @matthiasto about masking by looking for values ABOVE the threshold makes the most sense...

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
fso42: see my comment below

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,653 @@
...
+            updated dictionary
+        dtRangeTime: list
+            updated list of time steps
+    """
+
+    # load result type for fields
+    rangeTimeResType = cfgRangeTime['GENERAL']['rangeTimeResType']
+
+    # log info
+    log.info('Processing frame %d of time step: %.2f for range-time plot' % (len(mtiInfo['timeList'])+1, t))
```
fso42, line 569: Drop this to debug

fso42 (changes requested):

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,653 @@
...
+            array with x and y coordinates of radar location and point in direction of field of view
+
+    """
+
+    # fetch coordinate info
+    locationInfo = cfg['GENERAL']['radarLocation'].split('|')
+
+    # set radar location and point in direction of field of view as list
+    radarFov = np.asarray([[float(locationInfo[0]), float(locationInfo[1])], [float(locationInfo[2]),
+        float(locationInfo[3])]])
```
fso42, line 51: Please catch and add an log.error on non-existent location info in case of TTdiagram=false

fso42 (changes requested):

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
awirb: now we make a mask flowF > threshold and invert it

docs/moduleAna5Utils.rst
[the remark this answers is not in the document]
awirb: now we save the mtiInfo as pickle and this is available for plotting afterwards - so no need to save all the result fields but plotting is not inside com1DFA

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: treated now differently

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: same here

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
awirb: same heres

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: done

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: added an error if len(locationInfo) != 4

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+With the functions gathered in this module, flow variables of avalanche simulation results can be
+visualized in a distance versus time diagram, the so called **thalweg-time diagram**.
+The **tt-diagram** provides a way to identify main features of the temporal evolution of
+flow variables along the :term:`avalanche path`.
+This is based on the ideas presented in :cite:`FiFrGaSo2013` and :cite:`RaKo2020`, where
+avalanche simulation results have been transformed into the radar coordinate system to facilitate
+direct comparison, combined with the attempt to analyze simulation results in an avalanche path
+dependent coordinate system (:cite:`Fi2013`).
+In addition to the **tt-diagram**, :py:mod:`ana5RangeTime` also offers the possibility to
```
snowtechblog, line 18: I dont understand you naming conventions: pyhton distanceTimeAnalysis.py, here ana5RangeTime, in runscripts runRangeTimeDiagram.py or runThalwegTimeDiagram.py... to me a bit confusing

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+produce simulated **range-time diagrams** of the flow variables with respect to a radar field
+of view. With this, simulation results can be directly compared to radar measurements in terms
```
snowtechblog, line 20: after radar measurements add (for example MTI images from [ref])

ref = https://doi.org/10.1002/2017JF004375

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+of front position and inferred approach velocity. The colorcoding of the simulated
+**range-time** diagrams refers to average values of the chosen flow parameter
```
snowtechblog, line 22:
```suggestion
**range-time** diagrams are the average values of the chosen flow parameter
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+(e.g. flow depth, flow velocity) at specified range gates. This colorcoding is not directly
```
snowtechblog, line 23:
```suggestion
(e.g. flow depth FD, flow velocity FV) at specified range gates. This colorcoding is not directly
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+comparable to the intensity given in the range-time diagram from radar measurements.
```
snowtechblog, line 24:
```suggestion
comparable to the MTI intensity given in the range-time diagram from radar measurements.
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+
+.. Note::
+  The data processing for the **tt-diagram** and the **range-time diagram** can be done
+  *during run time* of :py:mod:`com1DFA`, or as a postprocessing step. However, the second option
+  requires saving all the required time steps of the flow variable fields, which is much more
```
snowtechblog, line 29:
```suggestion
  requires saving and later reading all the time steps of the flow variable fields, which is much more
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+  computationally expensive compared to the first option.
+
+To run
+~~~~~~~
+
+During run-time of :py:mod:`com1DFA`:
+
+* in your local copy of ``com1DFA/com1DFACfg.ini`` in [VISUALISATION] set `createRangeTimeDiagram`
+  to True and chose if you want a *TTdiagram* by setting this flag to True or in the case of a
```
snowtechblog, line 38:
```suggestion
  to True and choose if you want a *TTdiagram* by setting this flag to True or in the case of a
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+
+* have a look at :py:mod:`runScripts.runThalwegTimeDiagram.py` and :py:mod:`runScripts.runRangeTimeDiagram.py`
+
+* in your local copy of ``ana5Utils/distanceTimeAnalysisCfg.ini`` you can adjust the default settings
+  for the generation of the diagrams
+
+The resulting figures can be found in ``avalancheDirectory/Outputs/ana5Utils``.
+
+
+.. figure:: _static/thalwegTime_FD.png
```
snowtechblog, line 57: Approach velocity and value of max velocity have opaque white and not completely transparent background...

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+    :width: 90%
+
+    Thalweg-time diagram example.
```
snowtechblog, line 60:
```suggestion
    Thalweg-time diagram example: The y-axis contains the distance from the beta point along the avalanche path, e.g. the thalweg. Dots represent the avalanche front with the slope being the approach velocity. Red star marks the maximal approach velocity.
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+
+
+.. Note::
+  The **tt-diagram** requires info on an avalanche path (see :ref:`moduleAna3AIMEC:ana3AIMEC: Aimec`).
+  The simulated **range-time diagram** requires info on the x, y coordinate of the radar location, a point
```
snowtechblog, line 65:
```suggestion
  The simulated **range-time diagram** requires info on the x0, y0 coordinate of the radar location, a point xi, yi
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+  in the direction of the field of view, the aperture angle and the width of the range gates.
+  The maximum approach velocity indicated in the diagrams is computed as the ratio of the
```
snowtechblog, line 67:
```suggestion
  The maximum approach velocity is indicated in the distance-time diagrams with red star and is computed as the ratio of the
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+  distance travelled by the front and the respective time needed for two increments along the path/
+  radar's field of view.
```
snowtechblog, line 69:
```suggestion
  radar's field of view (e.g. this velocity over two range gates in fact contains a rolling average smoothing.).
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+
+
+Theory
+~~~~~~~~~
+
+Thalweg-time diagram
+~~~~~~~~~~~~~~~~~~~~~~
+
+First, the flow variable result field is transformed into a path-following coordinate system.
```
snowtechblog, line 78:
```suggestion
First, the flow variable result field is transformed into a path-following coordinate system of which the center line is the thalweg.
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+For this step, functions from :py:mod:`ana3AIMEC` are used.
+The distance to the *start of runout area point* is determined using a user defined threshold
+of the flow variable, giving the avalanche front position. The front positions defined with this
```
snowtechblog, line 81: I dont understand this sentence: *start of runout area point* is the beta point, right? What has it to do with the threshold to determine the avalanche (not only the front)?

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+method for all the time steps are shown as black dots in the **tt-diagram**.
+The mean values of the flow variable are computed at cross profiles along the avalanche path for
+each time step and included in the **tt-diagram** as colored field.
+For this analysis, all available flow variables can be chosen, but the interpretation of the
+tt-diagram structures and the corresponding meaning of avalanche front changes comparing flow thickness or flow velocity.
```
snowtechblog, line 86:
```suggestion
**tt-diagram** structures and the corresponding meaning of avalanche front may be different for flow thickness or flow velocity.
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Simulated Range-Time diagram
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The radar's field of view is determined using its location, a point in the direction of the field of
+view and the aperture angle. The distance of every point in the simulation results to the radar
```
snowtechblog, line 92:
```suggestion
view and the horizontal (azimuth) aperture angle of the antenna. The elevation or vertical aperture angle is not yet included. The line-of-sight distance of every grid point in the simulation results to the radar
```

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,105 @@
...
+location is computed. The simulation results which lie outside the radar's field of view are masked.
+The distance of the avalanche front with respect to the radar location is determined for a user
+defined threshold in the flow variable and the average values of range gates along the radar's line
+of sight are computed from the result field.
+This data is plotted in a range-time diagram, where the black dots indicate the avalanche front,
+and the colored field indicates the mean values of the flow parameter for the range gates for each
+time step.
```
snowtechblog, line 99:
```suggestion
location is computed. The simulation results which lie outside the radar's field of view are masked out.
The distance of the avalanche front with respect to the radar location is determined for a user
defined threshold in the flow variable and the average values of the result field for each range gates along the radar's line
of sight are computed.
This data is plotted in a range-time diagram, where the black dots indicate the avalanche front,
and the colored field indicates the mean values of the flow parameter for the range gates at each
time step.
```

snowtechblog (commented):
I finished with the documentation, but still want to run you code ...

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,730 @@
...
+        radarFov: numpy array
+            array with x and y coordinates of radar location and point in direction of field of view
+
+    """
+
+    # fetch coordinate info
+    locationInfo = cfg['GENERAL']['radarLocation'].split('|')
+
+    if len(locationInfo) != 4:
+        message = ('Radar location is not valid, required format x0|x1|y0|y1 not: %s' %
```
fso42, line 52:
```suggestion
        message = ('Radar location is invalid, required format x0|x1|y0|y1 not: %s' %
```

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
```diff
@@ -0,0 +1,50 @@
+### Config File - This file contains the main settings for the range time diagram in distanceTimeAnalysis
+## Copy to  local_distanceTimeAnalysisCfg.ini and set you parameters
+# This file is part of Avaframe.
+
+
+[GENERAL]
+# result type used for creating range time diagram e.g. FD, FV, ....
+rangeTimeResType = FD
+# time steps used for creating diagram
+tSteps = 0:1
```
fso42, line 10: Mh, maybe it would be better to rename this to something like 'distTimeTsteps` to avoid potential confusion with the tSteps from the com1DFA ini?

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
```diff
@@ -0,0 +1,50 @@
...
+
+# angle for the start of the run-out zone
+startOfRunoutAreaAngle = 10
+
+# chose interpolation method between 'nearest' and 'bilinear'
+interpMethod = bilinear
+
+# threshold distance [m]. When looking for the beta point make sure at least
+# dsMin meters after the beta point also have an angle bellow 10°
+dsMin = 30
```
fso42, line 43: These are all related to AIMEC, right?. Why not use the default configuration from that module instead of duplicating it here?

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
```diff
@@ -0,0 +1,50 @@
...
+
+
+[PLOTS]
+# every gateContours gate range plotted as contour line in radar field of view plot
+gateContours = 20
+# show debug plots
+debugPlot = False
```
fso42, line 50: take the debug plot flag from the main ini?

avaframe/in3Utils/geoTrans.py
```diff
@@ -831,3 +831,87 @@ def checkOverlap(toCheckRaster, refRaster, nameToCheck, nameRef, crop=False):
...
+
+def rotate(locationPoints, theta, deg=True):
+    """ rotate a vector provided as start and end point with theta angle
+
+        Parameters
+        -----------
+        locationPoints: list
+            list of lists with x,y coordinate of start and end point of a line
+        theta: float
+            rotation angle of the vector from start point to end point - degree default
```
fso42, line 878: Note rotatation direction. What happens with +theta or -theta values?

avaframe/in3Utils/geoTrans.py
```diff
@@ -831,3 +831,87 @@ def checkOverlap(toCheckRaster, refRaster, nameToCheck, nameRef, crop=False):
...
+
+    # create rotated line as list of start and end point
+    rotatedLine = [[locationPoints[0][0], float(locationPoints[0][0]+vectorRot[0])], # x
+           [locationPoints[1][0], float(locationPoints[1][0]+vectorRot[1])] #y
+          ]
+
+    # check if rotation worked correctly
+    lenVector = np.sqrt(vector[0]**2 + vector[1]**2)
+    lenVectorRot = np.sqrt(vectorRot[0]**2 + vectorRot[1]**2)
+    vecDiff = abs(lenVector - lenVectorRot)
```
fso42, line 913: Could this be a numpy.isclose or pytest.aprox application...?

fso42 (changes requested):

avaframe/ana3AIMEC/aimecTools.py
```diff
@@ -1059,3 +1064,200 @@ def getMaxMeanValues(rasterdataA, rasterArea):
...
+    rasterTransfo['xBetaPoint'] = rasterTransfo['x'][indStartOfRunout]
+    rasterTransfo['yBetaPoint'] = rasterTransfo['y'][indStartOfRunout]
+    rasterTransfo['startOfRunoutAreaAngle'] = angle[indStartOfRunout]
+
+    log.info('Start of run-out area at the %.2f ° point of coordinates (%.2f, %.2f)' %
+             (rasterTransfo['startOfRunoutAreaAngle'], rasterTransfo['xBetaPoint'], rasterTransfo['yBetaPoint']))
+
+    return rasterTransfo
+
+def makeDomainTransfoOnTheGo(avaDir, dem, cfgSetup, pathDict):
```
matthiasto, line 1188: Should we try to only have one  makeDomainTransfo? should I tackle this in an other PR?

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,730 @@
...
+
+    # fetch mask where result field not above threshold
+    maskAva = np.ma.masked_where(~(flowF > threshold), flowF)
+    bmaskAva = np.ma.getmask(maskAva)
+    # fetch mask of radar field of view
+    maskPhi = np.ma.getmask(rangeMasked)
+
+    # merge masks of result field not above threshold and outside radar field of view
+    maskFull = ~np.logical_and(~maskPhi, ~bmaskAva)
+
```
matthiasto, line 239: we do the same here and in extractMeanValuesAtRangeGates. Could we try to do it only once?

avaframe/com1DFA/com1DFA.py
```diff
@@ -1370,9 +1391,47 @@ def DFAIterate(cfg, particles, fields, dem):
         infoDict.update({'stopInfo': {'Stop criterion': '< %.2f percent of PKE' % stopCritPer,
                                       'Avalanche run time [s]': '%.2f' % avaTime}})
 
+    ######## create range time diagram #####################
+    # export data for range-time diagram
+    if cfg['VISUALISATION'].getboolean('createRangeTimeDiagram'):
```
matthiasto, line 1396: is cfg['VISUALISATION'] the good place for createRangeTimeDiagram? or should we have another section (I dont know how to call it) with createRangeTimeDiagram and createAvalanchPath?

avaframe/in3Utils/fileHandlerUtils.py
```diff
@@ -464,6 +464,36 @@ def exportcom1DFAOrigOutput(avaDir, cfg='', addTSteps=False):
     shutil.copy2(pathlib.Path('%s' % inputDir, 'ExpLog.txt'), outDir)
 
 
+def fetchFlowFields(flowFieldsDir, suffix=''):
```
matthiasto, line 467: is readFields not doing the same?

avaframe/in3Utils/geoTrans.py
```diff
@@ -831,3 +831,87 @@ def checkOverlap(toCheckRaster, refRaster, nameToCheck, nameRef, crop=False):
...
+    if vecDiff > 1.e-7:
+        log.warning('rotated vectors are not of same length %.2f vs %.2f' % (lenVector, lenVectorRot))
```
matthiasto, line 915: This should be in pytest, not in the function itself

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -0,0 +1,730 @@
...
+    # load lists
+    timeList = mtiInfo['timeList']
+    rangeList = mtiInfo['rangeList']
+
+    # as these might be not in ascending order timestep wise - order first
+    timeListSorted = np.asarray(timeList)[np.argsort(timeList)]
+    # convert to only positive distance values (move reference point to end of ava)
+    # to get correct distance increments for velocity computation
+    # nanmin required as if FV is zero range is nan
+    rangeListSorted = np.asarray(rangeList)[np.argsort(timeList)] + abs(np.nanmin(np.asarray(rangeList)))
```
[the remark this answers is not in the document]
awirb: this only works if avalanche simulations have been run before - it needs the fields to be available for the analysis, I add a sentence to the docu mentioning this, thanks for the hint!

docs/moduleAna5Utils.rst
awirb: thanks, I missed this one, should be ana5Utils

matthiasto (changes requested):

docs/moduleAna5Utils.rst
awirb: added

docs/moduleAna5Utils.rst
awirb: what do you think about show?

docs/moduleAna5Utils.rst
awirb: done

docs/moduleAna5Utils.rst
awirb: done

docs/moduleAna5Utils.rst
awirb: yep this is on purpose to make it 'more' visible but not fully hide the underlying lines..

docs/moduleAna5Utils.rst
awirb: so it can be for flow depth 1m  or 5cm or just >0 so that it everywhere where there is avalanche flow

docs/moduleAna5Utils.rst
awirb: included

docs/moduleAna5Utils.rst
awirb: adjusted

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
awirb: changed to distancetSteps

avaframe/in3Utils/geoTrans.py
awirb: added in counterclockwise

avaframe/ana3AIMEC/aimecTools.py
awirb: this would be great! - I can reference it in a n issue

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: again here if we only use one of the two functions, it won't be available but I can put this in a function and call it from both

avaframe/in3Utils/fileHandlerUtils.py
awirb: readFields really reads the data of the ascii files, here I just want to have the list of file paths - I moved it to fileHandlerUtils

avaframe/in3Utils/geoTrans.py
awirb: done

avaframe/com1DFA/com1DFA.py
awirb: good point, maybe tackle this in a future PR? when we came up with a good name

avaframe/in3Utils/geoTrans.py
matthiasto: I mean you know the points before and after the rotation so it should be possible to move lines 911-915 as well as recompute l 893 in the pytest

avaframe/ana5Utils/distanceTimeAnalysis.py
matthiasto: but you always want to compute the distance to the front AND then extract the mean values no?

awirb: so I made a function maskRangeFull that does these steps and is called from both, so that they can still be used individuall

avaframe/in3Utils/geoTrans.py
awirb: but I build them there again and do the check in the pytest

avaframe/in3Utils/geoTrans.py
awirb: changed it to use np.isclose(lenVector, lenVectorRot) here and moved to pytest

avaframe/in3Utils/geoTrans.py
matthiasto: Then I would suggest to remove it from the function here. I am not sure yhy you want to check that the function works here, we have pytests for this no?

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: made a function that is now called

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: maybe not always both, in case you are just interested in the distance?

avaframe/in3Utils/geoTrans.py
awirb: I moved it to pytests - and there just build once the vectors again

matthiasto: :+1:

avaframe/ana5Utils/distanceTimeAnalysis.py
matthiasto: Ok so now you put the mask stuff in a function, this is good but you still call the function twice... I would try to call the function outside and give the mask as input to minRangeSimulation and extractMeanValuesAtRangeGates.
My reasoning here is that if you say the ttdiagram thing is slow, we should avoid computing several times the same stuff.
Although I think what is slow is more the loop on the gates to get the mean value

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
```diff
@@ -0,0 +1,50 @@
...
+
+#+++++range time diagram with radar field of view +++++
+# width of range gates in meters
+rgWidth = 5.
+# aperture angle in degrees
+aperture = 15
+
+# radar location field of view points min (x0, y0) and point giving direction (x1, y1)
+# format: x0|x1|y0|y1 needs to be in same projection as DEM
+radarLocation =
```
snowtechblog, line 25: Add a check, if the radar location is within the valid area of the DEM. And tell the user, that the DEM is not suited for this radar location. Perhaps help him with generating the radarFieldOfViewPlotDEM (which does not need a z-value and any x,y coordinate value outside the DEM should work) already.

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,116 @@
...
+
+* in your local copy of ``com1DFA/com1DFACfg.ini`` in [VISUALISATION] set `createRangeTimeDiagram`
+  to True and choose if you want a *TTdiagram* by setting this flag to True or in the case of a
+  simulated range-time diagram to False
+
+* in your local copy of ``ana5Utils/distanceTimeAnalysisCfg.ini`` you can adjust the default settings
+  for the generation of the diagrams
+
+* run  :py:mod:`runScripts.runThalwegTimeDiagram.py` or :py:mod:`runScripts.runRangeTimeDiagram.py`
+  and set the `preProcessedData` flag to `True`
```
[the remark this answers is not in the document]
awirb: great thanks

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
awirb: an error occurs with message: 'Radar location is invalid, required format x0|x1|y0|y1 not: %s' %
            cfg['GENERAL']['radarLocation'])

avaframe/ana5Utils/distanceTimeAnalysis.py
awirb: Ok now I shortened this drastically, if the fully masked range array is available it is actually only one line - so I included this in the extract mean values function - so calling the mask function only once now and removed two other functions

avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -673,6 +620,12 @@ def importMTIData(avaDir, modName, inputDir='', simHash=''):
     else:
         mtiInfoPickles = list(inputDir.glob('*%s.p' % simHash))
 
+    if len(mtiInfoPickles) == 0:
+        message = ('No mtiInfo dictionary found in %s - consider first running avalanche simulations' %
+            inputDir)
+        log.error(message)
+        raise FileNotFoundError(message)
```
matthiasto, line 627: maybe use the function you just created in fU : fileNotFoundMessage

awirb: done

avaframe/runScripts/runRangeTimeDiagram.py
```diff
@@ -78,6 +80,15 @@
...
     demInputs = gI.readDEM(avalancheDir)
 
+    # fetch all flow parameter result fields
+    configDir = pathlib.Path(avalancheDir, 'Outputs', 'com1DFA', 'configurationFiles')
+    if (configDir.is_dir() is False):
+        fU.fileNotFoundMessage(('No configuration files found in %s - consider first running avalanche simulations' %
+            configDir))
+    elif len(list(configDir.glob('*.ini'))) == 0:
+        fU.fileNotFoundMessage(('No configuration files found in %s - consider first running avalanche simulations' %
+            configDir))
```
fso42, line 90: Twice exactly the same lines, shouldn't the checking be as simple or?

avaframe/runScripts/runThalwegTimeDiagram.py
```diff
@@ -73,6 +73,15 @@
     # fetch dem data
     demInputs = gI.readDEM(avalancheDir)
 
+    # fetch all flow parameter result fields
+    configDir = pathlib.Path(avalancheDir, 'Outputs', 'com1DFA', 'configurationFiles')
+    if (configDir.is_dir() is False):
+        fU.fileNotFoundMessage(('No configuration files found in %s - consider first running avalanche simulations' %
+            configDir))
+    elif len(list(configDir.glob('*.ini'))) == 0:
+        fU.fileNotFoundMessage(('No configuration files found in %s - consider first running avalanche simulations' %
```
fso42, line 82: same here

fso42 (approved):

avaframe/runScripts/runRangeTimeDiagram.py
awirb: true

avaframe/runScripts/runThalwegTimeDiagram.py
awirb: changed

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
awirb: if (np.any(radarFov[0] < np.amin(X)) or np.any(radarFov[0] > np.amax(X))  or
        np.any(radarFov[1] < np.amin(Y))  or np.any(radarFov[1] > np.amax(Y))):
        message = 'Radar location outside of DEM, x= %.2f, %.2f, y= %.2f, %.2f' % (radarFov[0][0],
            radarFov[0][1], radarFov[1][0], radarFov[1][1])
        log.error(message)
        raise AssertionError(message)

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
awirb: decided to keep this

docs/moduleAna5Utils.rst
```diff
@@ -0,0 +1,116 @@
...
+~~~~~~~~~
+
+Thalweg-time diagram
+~~~~~~~~~~~~~~~~~~~~~~
+
+First, the flow variable result field is transformed into a path-following coordinate system, of
+which the centerline is the :term:`avalanche path`.
+For this step, functions from :py:mod:`ana3AIMEC` are used.
+The distance to the *start of runout area point* is determined using a user defined threshold
+of the flow variable, giving the avalanche front position. The front positions defined with this
```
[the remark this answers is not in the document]
awirb: it is not masked, it is just checking where is the value above threshold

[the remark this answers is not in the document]
awirb: I tried to make this clearer in the sentence

matthiasto (approved):

jtfisch (approved):

fso42 (approved):

fso42 (approved):

fso42 (approved):

PR #693, opened 2022-03-18
Alpha Beta optional in runOperational

fso42:
- make AB optional in runOperational (related to avaframeconnector)
- (hacky) solution to handle apostrophes in filenames

### DIFF
--- a/avaframe/runOperational.py
+++ b/avaframe/runOperational.py
@@ -77,17 +77,27 @@ def runOperational(avalancheDir=''):
     inputDir = avaDir / 'Outputs' / 'com1DFA' / 'peakFiles'
     peakFilesDF = fU.makeSimDF(inputDir, avaDir=avaDir)
 
-    # Run Alpha Beta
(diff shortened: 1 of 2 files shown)


avaframe/runOperational.py
```diff
@@ -77,18 +77,28 @@ def runOperational(avalancheDir=''):
...
+
+        # if not: Run Alpha Beta
+        cfgAB = cfgUtils.getModuleConfig(com2AB)
+        resAB = com2AB.com2ABMain(cfgAB, avalancheDir)
+        abShpFile = outAB.writeABtoSHP(resAB)
+
+        # Generate report info for com2AB
+        reportDictList, _, _ = outAB.writeABpostOut(resAB, cfgAB, reportDictList)
+
+    except AssertionError as e:
```
matthiasto, line 93: are we sure AssertionError is the only error that might pop out?

awirb (approved):

Issue #715, opened 2022-04-28
User Feedback (CT)


fso42:
- [x] Operational run: option to not delete directory, but insert 
- [x] Have info about variation in file name / attributes (for easier overview in QGis...) -> maybe experimental qgis script?!
- [x] Difference from ini defaults should also go to the report

fso42:
Done with PR #750

PR #717, opened 2022-05-05
Move sim hash

### LINKED ISSUE #690: Move simHash in filename
To ensure compatibility with existing simulation workflows, moving the simHash would help:
from 
XXXX_dfa_SimHASH_ppr.csv
to
XXXX_SimHASH_dfa_ppr.csv

fso42:
- change order of simHash; fixes #690 
- set deleteOutput to False in runOperational; addresses #715


avaframe/ana5Utils/__init__.py
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
matthiasto, line 2: What do we need this for?

matthiasto, line 2: What is this needed for?

PR #738, opened 2022-07-06
fix particle exiting domain and com3Hybrid

matthiasto:
Particle exiting domain was not working properly with the new cython flag thing (wraparound boundcheck...)
This is fixed and the hybrid Path ini file is now up to date

### DIFF
--- a/avaframe/com1DFA/DFAfunctionsCython.pyx
+++ b/avaframe/com1DFA/DFAfunctionsCython.pyx
@@ -10,7 +10,6 @@
 import copy
 import logging
 import math
-import cython
 import numpy as np
 cimport numpy as np
 from libc cimport math as math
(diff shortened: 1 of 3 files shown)


avaframe/com1DFA/DFAfunctionsCython.pyx
```diff
@@ -1879,7 +1879,7 @@ cpdef (double, double, int, int, int, double, double, double, double) samosProje
   iCell = getCells(xNew, yNew, ncols, nrows, csz)
   if iCell < 0:
     # if not on the DEM exit with iCell=-1
```
awirb, line 1881: do you need to update the comment then too? maybe add explanation

fso42 (approved):

PR #742, opened 2022-07-12
Use SecRel by default

matthiasto:
- Change logic for secondary release areas:
  - Check for release and secondary release shapefiles
  - If only release available -> just run release
  - If both release and secondary release are available -> run with secondary release UNLESS secRelArea = False
ALL scenarios in release get the secondary release areas!


avaframe/in3Utils/geoTrans.py
```diff
@@ -181,87 +181,81 @@ def resizeData(raster, rasterRef):
         return raster['rasterData'], rasterRef['rasterData']
 
 
-def getMeshXY(rasterDict, cellSizeNew=None):
-    """ Get x and y vectors for mesh with given number of rows and columns, lowerleftcenter and cellSize
+def remeshData(rasterDict, cellSizeNew, remeshOption='griddata', interpMethod='cubic', larger=True):
```
fso42, line 184: please add interpMethod to the docstring

avaframe/in3Utils/geoTrans.py
```diff
@@ -286,55 +280,23 @@ def remeshDEM(demFile, cfgSim):
...
-    headerRemeshed['noDataValue'] = headerDEM['noDataValue']
-
-    # write new DEM dictionary
-    remeshedDEM = {'header': headerRemeshed}
-    xNewGrid, yNewGrid = np.meshgrid(xNew, yNew)
-    zNew = sp.interpolate.griddata((xGrid, yGrid), z, (xNewGrid, yNewGrid), method='cubic',
-                                   fill_value=headerDEM['noDataValue'])
-    log.info('Remeshed data extent difference x: %f and y %f' % (diffExtentX, diffExtentY))
-    remeshedDEM['rasterData'] = zNew
+    log.info('Remeshing the input DEM (of cell size %.4g m) to a cell size of %.4g m' % (cszDEM, cszDEMNew))
```
fso42, line 298: I would change it to .2g instead of .4g. This is already plenty enough...

avaframe/in3Utils/geoTrans.py
```diff
@@ -938,27 +899,77 @@ def rotate(locationPoints, theta, deg=True):
...
     rotatedLine = [[locationPoints[0][0], float(locationPoints[0][0]+vectorRot[0])], # x
-           [locationPoints[1][0], float(locationPoints[1][0]+vectorRot[1])] #y
-          ]
+                   [locationPoints[1][0], float(locationPoints[1][0]+vectorRot[1])] #y
+                   ]
 
     return rotatedLine
 
 
+def makeCoordGridFromHeader(rasterHeader, cellSizeNew=None, larger=True):
```
fso42, line 908: Please put larger in the docstring

fso42 (changes requested):

avaframe/in3Utils/geoTrans.py
```diff
@@ -181,87 +181,83 @@ def resizeData(raster, rasterRef):
...
-
-
-def remeshData(rasterFile, cellSize):
-    """ compute raster data on a new mesh with cellSize using scipy RectBivariateSpline
-
-        the new mesh is as big or smaller as the original mesh
+        remeshOption are to choose between 'griddata', 'interp2d' or 'RectBivariateSpline'
+        Only the 'griddata' works properly if the input data contains noData points,
+        'interp2d' or 'RectBivariateSpline' are faster but fail if input data contains noData points.
+        The new mesh is as big or smaller as the original mesh
```
awirb, line 190: or bigger if larger=True

avaframe/in3Utils/geoTrans.py
```diff
@@ -181,87 +181,83 @@ def resizeData(raster, rasterRef):
...
-        return x, y, xExtent, yExtent
-
-
-def remeshData(rasterFile, cellSize):
-    """ compute raster data on a new mesh with cellSize using scipy RectBivariateSpline
-
-        the new mesh is as big or smaller as the original mesh
+        remeshOption are to choose between 'griddata', 'interp2d' or 'RectBivariateSpline'
+        Only the 'griddata' works properly if the input data contains noData points,
+        'interp2d' or 'RectBivariateSpline' are faster but fail if input data contains noData points.
```
awirb, line 189: does this throw an error?

avaframe/in3Utils/geoTrans.py
```diff
@@ -380,7 +344,7 @@ def searchRemeshedDEM(demName, cfgSim):
 
     # fetch info on desired meshCellSize
     meshCellSize = float(cfgSim['GENERAL']['meshCellSize'])
-    meshCellSizeThreshold = float( cfgSim['GENERAL']['meshCellSizeThreshold'])
+    meshCellSizeThreshold = float(cfgSim['GENERAL']['meshCellSizeThreshold'])
```
awirb, line 347: consider using getfloat here

awirb (changes requested):

avaframe/com1DFA/com1DFA.py
```diff
@@ -825,7 +807,7 @@ def initializeSimulation(cfg, demOri, inputSimLines, logName):
     entrMassRaster = geoTrans.checkOverlap(entrMassRaster, relRaster, 'Entrainment', 'Release', crop=True)
     # check for overlap with the secondary release area
     if secondaryReleaseInfo['flagSecondaryRelease'] == 'Yes':
-        for secRelRaster in secondaryReleaseInfo['rasterList']:
+        for secRelRaster in secondaryReleaseInfo['rasterData']:
```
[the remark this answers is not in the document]
matthiasto: Ok, I found out why I changed this, the function prepareArea updated the dictionary with the 'rasterData' key which actually can be a list of rasters. If I name the 'rasterList' key 'rasterList', then I have in my dictionary once 'rasterList' and once  'rasterData' with almost exactly the same data and I do not need the  'rasterData' key anymore. So I could decide to keep the 'rasterList' key and delete the 'rasterData'  one, or I just overwrite the 'rasterData'  key. This is what I did but I can change it if not suited.

avaframe/com1DFA/com1DFA.py
```diff
@@ -1126,6 +1108,56 @@ def initializeFields(cfg, dem, particles):
...
+        # remove overlap with main release areas
+        noOverlaprasterList = []
+        for secRelRatser, secRelName in zip(secondaryReleaseInfo['rasterData'], secondaryReleaseInfo['Name']):
+            noOverlaprasterList.append(geoTrans.checkOverlap(secRelRatser, relRaster, 'Secondary release ' + secRelName,
+                                                             'Release', crop=True))
+
+        secondaryReleaseInfo['flagSecondaryRelease'] = 'Yes'
+        # replace the rasterData with noOverlaprasterList (which is the list of rasterData without the overlapping
+        # part with the release)
+        secondaryReleaseInfo['rasterData'] = noOverlaprasterList
```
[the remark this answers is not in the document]
matthiasto: Ok, I found out why I changed this, the function prepareArea updated the dictionary with the 'rasterData' key which actually can be a list of rasters. If I name the 'rasterList' key 'rasterList', then I have in my dictionary once 'rasterList' and once  'rasterData' with almost exactly the same data and I do not need the  'rasterData' key anymore. So I could decide to keep the 'rasterList' key and delete the 'rasterData'  one, or I just overwrite the 'rasterData'  key. This is what I did but I can change it if not suited.

avaframe/in3Utils/geoTrans.py
matthiasto: No it does not but the default option is 'griddata' and it is the one used in com1DFA. Should it check for nans if it is used with the other methods?

avaframe/com1DFA/com1DFA.py
```diff
@@ -495,7 +495,7 @@ def prepareInputData(inputSimFiles, cfg):
         if entResInfo['flagSecondaryRelease'] == 'Yes':
             secondaryReleaseFile = inputSimFiles['secondaryReleaseFile']
             secondaryReleaseLine = shpConv.readLine(secondaryReleaseFile, '', demOri)
-            secondaryReleaseLine['fileName'] = [secondaryReleaseFile]
+            secondaryReleaseLine['fileName'] = secondaryReleaseFile
```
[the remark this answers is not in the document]
matthiasto: it is then used in secondaryReleaseInfo['fileName'] for the log.info. It make not difference if it is a list or a string

avaframe/in3Utils/geoTrans.py
awirb: If the methods do not work properly in this case and hence should not be used, I think it is a good idea to first check for nans and if they are there either use griddata and throw a warning or throw directly an error and say that the chosen method doesn't work as there are nans if this is the case

awirb (changes requested):
consider adding a SECREL for one of the generic avas, so that there is an example available on the example avas and add a note in the docu about this behaviour?

fso42 (approved):

PR #743, opened 2022-07-26
Update gettingstarted.rst

matthiasto:


awirb (approved):
or as this is almost a main functionality should we consider moving it outside?

fso42 (approved):

PR #746, opened 2022-07-28
Bugfixes in rangeTimePlots (contr. Anselm K.)

snowtechblog:
Hi Avaframe Team,

Thank you for the great simulation tool, I am really impressed how easy to use the tool is as soon one got a little starting help (thanks @matthiasto).

I had a look at the rangeTimePlots for the thalweg-time-diagrams (thanks @awirb for the great work), and fixed some bugs. Namely:
- the small velocity legend did not react to zoom/pan events of the figure, e.g. as soon as one programmatically changes the xlim or ylim of the plot, the legend was wrong.
- the max of the front velocity suffered some inconsistencies, thus, I tried to find a way how to neglect obviously (to large) velocities. For example, I found one case where the were 400m/s in the first second inside the release area... The first commit 3e516ad is overwritten from the last commit 4f5b9c8, however, the way the max front velocity was calculated before is still in the code and can be activated with the flag `old_algo=True` in `ana5Utils.distanceTimeAnalysis.approachVelocity`

Other small changes include a found double import, explicit setting of mpl backend 'agg' in some debugPlots function, and some PyCharm files were not excluded in the gitignore file.

Here 2 plots that show the correction of the velocity legend, with the following steps
- make range-time-plot with legend filling the full axes
- zoom to 0-100s and -1000m-0m (from beta point)
- now: diagonal should equate to 10m/s

Not working: zoom does not change the velocity legend...
thalwegTime_FV_e557d4d8f9_wrong_legend_after_zooming

Working: zoom causes velocity legend to update and shows the expected result fro the diagonal velocity:
thalwegTime_FV_e557d4d8f9_correct_legend_after_zooming

And here 2 plots that show the updated way to calculate the max front velocity:

Not working: Max velocity found at first timestep somewhere in release area:
thalwegTime_FV_e557d4d8f9_old_algo_True

Working: max velocity found in the steepest part of the flowing front:
thalwegTime_FV_e557d4d8f9_correct_maxvel

Note: the small kink in the front at 30s and -1300m distance is also neglected. We should discuss how strong the max front algorithm should filter those kinks.

Cheers, Anselm

### DIFF
--- a/avaframe/ana5Utils/distanceTimeAnalysis.py
+++ b/avaframe/ana5Utils/distanceTimeAnalysis.py
@@ -6,21 +6,15 @@
 
 
 import numpy as np
-import matplotlib as mpl
-from matplotlib import pyplot as plt
 import logging
 import pickle
 import pathlib
-import configparser
 
 # Local imports
 from avaframe.in3Utils import cfgUtils
-from avaframe.in3Utils import logUtils
-import avaframe.in2Trans.ascUtils as IOf
 import avaframe.in3Utils.geoTrans as gT
 import avaframe.out3Plot.outDistanceTimeAnalysis as dtAnaPlots
 import avaframe.ana3AIMEC.aimecTools as aT
-import avaframe.out3Plot.plotUtils as pU
 import avaframe.com1DFA.com1DFA as com1DFA
 import avaframe.in3Utils.fileHandlerUtils as fU
 
@@ -238,7 +232,7 @@ def extractFrontAndMeanValuesRadar(cfgRangeTime, flowF, mtiInfo):
     # mask range with radar field of view and treshold of flow variable result
     maskAva, bmaskAvaRadar, rMaskedAvaRadar = maskRangeFull(flowF, threshold, rangeMasked)
 
-    #++++++++Eextract front location with respect to radar ++++++++++++++
+    # ++++++++Extract front location with respect to radar ++++++++++++++
     # get line of sight distance to identify front, use threshold to mask flowF
     # line of sight min of masked array
     losDistance = rMaskedAvaRadar.min()
@@ -246,7 +240,7 @@ def extractFrontAndMeanValuesRadar(cfgRangeTime, flowF, mtiInfo):
     # update lists of time step and front location
     mtiInfo['rangeList'].append(losDistance)
 
-    #+++++++Extract average values at range gates +++++++++
+    # +++++++Extract average values at range gates +++++++++
     # min and max radar range of masked radar range
     if not rMaskedAvaRadar.mask.all():
         #TODO why int?
@@ -438,7 +432,6 @@ def extractFrontAndMeanValuesTT(cfgRangeTime, flowF, demHeader, mtiInfo):
     # TODO: average over cells – weighted with cell area (aimec function)
     rasterArea = rasterTransfo['rasterArea']
     maxaCrossMax, aCrossMax, aCrossMean = aT.getMaxMeanValues(slRaster, rasterArea)
-
     # use the max or the mean of each cross section
     if cfgRangeTime['GENERAL']['maxOrMean'].lower() == 'max':
         aCross = aCrossMax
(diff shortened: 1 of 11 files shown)


avaframe/ana5Utils/distanceTimeAnalysis.py
```diff
@@ -696,16 +713,55 @@ def approachVelocity(mtiInfo, minVelTimeStep):
...
+        nan_idx = np.argwhere(~np.isnan(rangeListSorted)).squeeze()
+        # remove anything behind runout
+        rmax_idx = np.arange(0,np.nanargmax(rangeListSorted)+1)
+        idx = np.intersect1d(nan_idx, rmax_idx)
+
+        rangeListSorted_small = rangeListSorted[idx]
+        timeListSorted_small = timeListSorted[idx]
+        # remove non-unique range values
+        rangeListSorted_unique, unique_idx = np.unique(rangeListSorted_small.round(decimals=2), return_index=True)
+        timeListSorted_unique = timeListSorted_small[unique_idx]
```
matthiasto, line 739: Why do you need to remove these values?

matthiasto (changes requested):

avaframe/ana5Utils/distanceTimeAnalysisCfg.ini
```diff
@@ -15,6 +15,8 @@ distanceTimeSteps = 0:1
 thresholdResult = 1.e-5
 # set minimum time step difference for computing approach velocity > 2s for discarding spurious effects
 minVelTimeStep = 2.
+# use the maximum rangeTimeResType or the mean rangeTimeResType (max or mean)
+maxOrMean = max
```
awirb, line 19: this is only applied for the mean or max values of the TT diagram not the range time with radar field of view - so either this should be moved to the tt section of the configuration settings or also implemented for the range time diagram function: extractFrontAndMeanValuesRadar

avaframe/out3Plot/outDistanceTimeAnalysis.py
```diff
@@ -284,9 +292,127 @@ def plotMaskForMTI(cfgRangeTime, bmaskRange, bmaskAvaRadar, bmaskAvaRadarRangesl
     pU.saveAndOrPlot(pathDict, outFileName, fig)
 
 
-def rangeTimeVelocityLegend(timeSteps, rangeGates, ax, maxVel, width=0.25, height=0.25, lw=0.5, textsize=7):
-    """ set legend in range time diagram for velocity in terms of steepness of front position
+def rangeTimeVelocityLegend(ax, maxVel, width=0.25, height=0.25, lw=0.5, textsize=7):
```
fso42, line 295: see my general remark about variable naming within this function

avaframe/out3Plot/outDistanceTimeAnalysis.py
```diff
@@ -284,9 +292,127 @@ def plotMaskForMTI(cfgRangeTime, bmaskRange, bmaskAvaRadar, bmaskAvaRadarRangesl
...
+    ----------
+    timesteps : ndarray
+        Array containing the timesteps in [s]
+    rangegates : ndarray
+        Array containing the rangegates in [m]
+    ax : axes object
+        Axes object where to put legend in
+    width : float , *0.25
+        fractional percentage of legend width
+    '''
```
fso42, line 309: These do not match the actual function parameters, please update.

avaframe/out3Plot/outDistanceTimeAnalysis.py
```diff
@@ -284,9 +292,127 @@ def plotMaskForMTI(cfgRangeTime, bmaskRange, bmaskAvaRadar, bmaskAvaRadarRangesl
...
+
+    # make the text, e.g. the mandatory first=True call
+    zoom_lvl_change(ax, first=True)
+
+    # connect callback
+    ax.callbacks.connect('xlim_changed', zoom_lvl_change)
+    ax.callbacks.connect('ylim_changed', zoom_lvl_change)
+
+
+def rangeTimeVelocityLegend_depreciated(timeSteps, rangeGates, ax, maxVel, width=0.25, height=0.25, lw=0.5, textsize=6):
```
fso42, line 414: I assume this is meant to be 'deprecated'? 
If so, then there's no point in keeping it, so just remove these functions

avaframe/out3Plot/outDistanceTimeAnalysis.py
```diff
@@ -405,3 +530,44 @@ def getVelocityPoints(xMax, yMax, width, height, xinterval, yinterval, maxVel, k
...
+        Note: used to be an internal function to rangeTimeVelocityLegend,
+              so use with care.
+
+        Parameters
+        -----------
+        xMax: float
+            max x extent of data (timeSteps)
+        yMax: float
+            max y extent of data (rangeGates)
+
```
fso42, line 554: please add all input parameters here

fso42 (changes requested):
Hi @snowtechblog, thanks for your contribution; it is much appreciated!
Two general points:
- I'm not sure whether I am missing something, but the first two plots in your comment are the same, right?
- We have the coding style of using camelCase for variable naming (and no underscores). I now put this up in the notes to developers in the documentation; see https://docs.avaframe.org/en/latest/develop.html. Since this wasn't up at the time of your contribution, if you want, I can change the source files and send them to you so you can commit them and get the correct attribution.

snowtechblog:
>     * I'm not sure whether I am missing something, but the first two plots in your comment are the same, right?

No, they are actually different: See the velocity legend (velocity sun) values. In the first plot, the diagonal has still 6.3m/s value, but it must update to the new zoom ratio which gives then 10m/s as shown in the second plot

awirb (changes requested):

fso42 (approved):

PR #752, opened 2022-08-19
Removes triple printout; fixes #673

### LINKED ISSUE #673: for now running com1DFA - configuration is printed to log several times

fso42:
Removes multiple printouts of config during run in, e.g. com1DFA
- makes compareConfig always honour the toPrint flag
- However, ALWAYS logs the differences!
- Improve operational installation instructions
- Add german version of operational installation


matthiasto (changes requested):
Why are we doing twice the same thing in the beginning?
cfgUtils - INFO -  Reading config from: /home/matthias/Documents/github/AvaFrame/avaframe/com1DFA/com1DFACfg.ini and /home/matthias/Documents/github/AvaFrame/avaframe/com1DFA/local_com1DFACfg.ini
Could we do this only once or at least log the difference only once?

fso42:
@awirb: regarding @matthiasto comment: 
Both the functions 'setRelThIni' and 'getThickness' can override the initial cfg object with some cfgFile again. We could completely eliminate this since the override is already done in the first step in com1DFA Main as soon as cfgStart is read (where the override would be applied), right? 
So we have the two functions just accept a cfgParser object instead of a cfgParser obj + override file, and we note that any overrides need to be done before that, which is what we want anyways. Or is there a compelling reason not to do this?

PR #753, opened 2022-08-22
Add normal distribution from mean and ci

awirb:
* add variation option for thickness settings and probrun based on normal distribution derived from ci and mean 
* fix bugs in deriving variation
*  add filtering option to aimec

### DIFF
--- a/avaframe/ana3AIMEC/aimecTools.py
+++ b/avaframe/ana3AIMEC/aimecTools.py
@@ -87,7 +87,7 @@ def readAIMECinputs(avalancheDir, pathDict, dirName='com1DFA'):
     return pathDict
 
 
-def fetchReferenceSimNo(avaDir, inputsDF, comModule, cfgSetup):
+def fetchReferenceSimNo(avaDir, inputsDF, comModule, cfg):
     """ Define reference simulation used for aimec analysis.
 
         if the configuration files are available and a varParList is provided, the simulations
@@ -105,7 +105,7 @@ def fetchReferenceSimNo(avaDir, inputsDF, comModule, cfgSetup):
             dataFrame with simulations to analyze and path to result files
         comModule: str
             computational module used to produce the results to analyze
-        cfgSetup: configParser object
+        cfg: configParser object
             configuration for aimec - referenceSimValue, varParList used here
 
         Returns
@@ -121,6 +121,8 @@ def fetchReferenceSimNo(avaDir, inputsDF, comModule, cfgSetup):
         colorVariation: boolean
             True if a color variation should be applied in the plots
     """
+
+    cfgSetup = cfg['AIMECSETUP']
     inputDir = pathlib.Path(avaDir, 'Outputs', comModule, 'peakFiles')
     if not inputDir.is_dir():
         message = 'Input directory %s does not exist - check anaMod' % inputDir
@@ -144,6 +146,14 @@ def fetchReferenceSimNo(avaDir, inputsDF, comModule, cfgSetup):
             raise e(message)
         else:
             configFound = False
+
+    # filter simulations
+    if cfg.has_section('FILTER'):
+        parametersDict = fU.getFilterDict(cfg, 'FILTER')
+        simNameList = cfgHandling.filterSims(avaDir, parametersDict, specDir='')
+        inputsDF = inputsDF[inputsDF['simName'].isin(simNameList)]
+        log.info('%d simulations found matching filter criteria' % inputsDF.size[0])
+
     # if the simulations come from com1DFA, it is possible to order the files and define a reference
     if configFound and cfgSetup['varParList'] != '':
         # fetch parameters that shall be used for ordering
--- a/avaframe/ana3AIMEC/aimecTools.py
+++ b/avaframe/ana3AIMEC/aimecTools.py
@@ -152,7 +152,7 @@ def fetchReferenceSimNo(avaDir, inputsDF, comModule, cfg):
         parametersDict = fU.getFilterDict(cfg, 'FILTER')
         simNameList = cfgHandling.filterSims(avaDir, parametersDict, specDir='')
         inputsDF = inputsDF[inputsDF['simName'].isin(simNameList)]
-        log.info('%d simulations found matching filter criteria' % inputsDF.size[0])
+        log.info('%d simulations found matching filter criteria' % inputsDF.shape[0])
 
     # if the simulations come from com1DFA, it is possible to order the files and define a reference
     if configFound and cfgSetup['varParList'] != '':
--- a/avaframe/ana3AIMEC/ana3AIMECCfg.ini
+++ b/avaframe/ana3AIMEC/ana3AIMECCfg.ini
@@ -65,6 +65,13 @@ referenceSimName =
 unit =
 
 #---------------------------------------
+## Uncomment this section FILTER in your local copy of the ini file and add filter parameter and parameter values
+## see the example provided below for release thickness
+#[FILTER]
+## define parameter and corresponding values from the simulation configuration to filter simulations
+## multiple parameters are possible, just add them in a new line each
+##relTh = 0.75|0.9|0.8
+
 
 
 # plot save results flags------------------------
--- a/avaframe/ana4Stats/probAna.py
+++ b/avaframe/ana4Stats/probAna.py
@@ -159,8 +159,12 @@ def updateCfgRange(cfg, cfgProb, varName, varDict):
             valValues = fU.splitIniValueToArraySteps(cfg['GENERAL'][varName])
 
         elif cfgProb['PROBRUN']['variationType'].lower() == 'range':
-            valStart = str(float(valVal) - float(valVariation))
-            valStop = str(float(valVal) + float(valVariation))
+            if '-' in valVariation or '+' in valVariation:
+                valStart = str(float(valVal) + float(valVariation))
+                valStop = float(valVal)
+            else:
+                valStart = str(float(valVal) - float(valVariation))
+                valStop = str(float(valVal) + float(valVariation))
             cfg['GENERAL'][varName] = '%s:%s:%s' % (valStart, valStop, valSteps)
             valValues = np.linspace(float(valStart), float(valStop), int(valSteps))
         else:
--- a/avaframe/ana4Stats/probAna.py
+++ b/avaframe/ana4Stats/probAna.py
@@ -49,9 +49,9 @@ def createComModConfig(cfgProb, avaDir, modName, cfgFileMod=''):
     fU.makeADir(outDir)
 
     # loop over all parameters for performing parameter variation
-    varParList = cfgProb['PROBRUN']['varParList'].split('|')
+    variationsDict = makeDictFromVars(cfgProb['PROBRUN'])
     cfgFiles = {}
-    for varName in varParList:
+    for varName in variationsDict:
         # define configuration files
         # get filename of module
         modNameString = str(pathlib.Path(modName.__file__).stem)
@@ -59,7 +59,7 @@ def createComModConfig(cfgProb, avaDir, modName, cfgFileMod=''):
 
         # use cfgFile, local com module settings or default settings if local not available
         modCfg = cfgUtils.getModuleConfig(modName, fileOverride=cfgFileMod)
-        modCfg = updateCfgRange(modCfg, cfgProb, varName)
+        modCfg = updateCfgRange(modCfg, cfgProb, varName, variationsDict[varName])
         with open(cfgFile, 'w') as configfile:
             modCfg.write(configfile)
         # append cfgFiles to list
@@ -68,7 +68,7 @@ def createComModConfig(cfgProb, avaDir, modName, cfgFileMod=''):
     return cfgFiles
 
 
-def updateCfgRange(cfg, cfgProb, varName):
+def updateCfgRange(cfg, cfgProb, varName, varDict):
     """ update cfg with a range for parameters in cfgProb
 
         Parameters
@@ -79,6 +79,8 @@ def updateCfgRange(cfg, cfgProb, varName):
             configparser object with info on update
         varName: str
             name of parameter used for variation
+        varDict: dict
+            dictionary with variationValue and numberOfSteps for varName
 
         Returns
         --------
@@ -109,9 +111,10 @@ def updateCfgRange(cfg, cfgProb, varName):
                 log.error(message)
                 raise AssertionError(message)
 
+    # this is now done for parameter VARNAME from inputs
     # get range, steps and reference value of parameter to perform variations
-    valVariation = cfgProb['PROBRUN']['%sVariation' % varName]
-    valSteps = cfgProb['PROBRUN']['%sSteps' % varName]
+    valVariation = varDict['variationValue']
+    valSteps = varDict['numberOfSteps']
     valVal = cfg['GENERAL'][varName]
 
     # set variation in configuration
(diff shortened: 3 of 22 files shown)


avaframe/ana4Stats/probAnaCfg.ini
```diff
@@ -20,24 +20,40 @@ peakLim = 1.0
...
+# # varParValue is the reference value (which is read from computational module settings)
+# variationType = percent
+# muVariation = 35
+# relThVariation = 35
+# #++++++NUMBER OF STEPS
+# # how many steps in variation - if reference value not included will be appended
+# muSteps = 11
+# relThSteps = 11
+# #if reference values shall be added to variation set addStandardConfig to True
+# addStandardConfig = False
```
matthiasto, line 44: Does this need to be commented?

avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -69,6 +69,11 @@ relThPercentVariation =
 # if a variation on relTh shall be performed add here +- absolute value and number of steps separated by $
 # for example relThRangeVariation=0.5$10
 relThRangeVariation =
+# if variation on relTh shall be performed using a normal distribution in number of steps,
+# value of buildType (ci95 value), min and max of dist in percent, buildType (ci95 only allowed),
+# support (e.g. 10000) all separated by $: e.g. normaldistribution$numberOfSteps$0.3$95$ci95$10000
+# if relThFromShp=True ci95 is read from shp file too
+relThDistVariation =
```
matthiasto, line 76: why put normaldistribution at the beginning of normaldistribution$numberOfSteps$0.3$95$ci95$10000.
It is the only option no?

avaframe/in2Trans/shpConversion.py
```diff
@@ -140,6 +145,7 @@ def SHP2Array(infile, defname=None):
     SHPdata['y'] = Coordy
     SHPdata['z'] = Coordz
     SHPdata['id'] = idList
+    SHPdata['ci95'] = ci95List
```
matthiasto, line 148: add this to the docstring info

matthiasto (changes requested):
I think you will need to test the extract gaussian functions no?

avaframe/ana4Stats/probAnaCfg.ini
awirb: the problem is if it is not commented it is not possible to add other variation and step values for parameters that are not in the default settings, if the full Section is 'new', additional keys are allowed

avaframe/com1DFA/com1DFACfg.ini
awirb: so far yes, but already wanted to keep it flexibel if more distributions will be added

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -444,6 +488,19 @@ def setVariationForAllFeatures(cfg, key, thType, varType, variationFactor):
...
     # fetch thickness list
     thicknessList = cfg['INPUT'][thType + 'Thickness'].split('|')
+    # fetch ci95 list
+    ci95List = cfg['INPUT'][thType + 'Ci95'].split('|')
+
+    # do some preprocessing if varTypu is dist
+    if varType == 'Dist':
+        distInfo = variationFactor.split('$')
+        if len(distInfo) != 7:
+            message = 'Format of distVariation string is false'
```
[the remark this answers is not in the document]
awirb: Done

avaframe/in1Data/computeFromDistribution.py
```diff
@@ -120,3 +123,67 @@ def getEmpiricalCDFNEW(sample):
...
+        x: numpy array
+            support x
+    """
+
+    # load parameters from config
+    sampleSize = int(cfg['sampleSize'])
+    mean = float(cfg['mean'])
+    minMaxInterval = float(cfg['minMaxInterval'])
+
+    # first check if std or ci95 is provided throw error if both
```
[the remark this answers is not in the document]
awirb: a this comment is from the first version, now you have to provide the buildType before it was std and ci95 and you could leave one empty, but I added an error as else if it is none of the available options.

avaframe/in2Trans/shpConversion.py
awirb: done

avaframe/ana4Stats/probAnaCfg.ini
fso42: I agree here: I don't think a completely commented out section is a good idea, this should be solved differently then.

awirb: ok now solved differently varParList, variationValue and numberOfSteps each separated by | for each parameter

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -346,6 +358,24 @@ def splitVariationToArraySteps(value, key, fullCfg):
...
+        # if not already appended the step of the distribution values that shall be taken as value
+        # add to string so it is clear which value shall be taken from distribution
+        # first check format of string
+        if len(itemsL) == 6:
+            if 'distribution' in itemsL[0]:
+                for i in range(int(itemsL[1])):
+                    itemsArray.append('%d$' % i + value)
+            else:
+                message = ('Format of %s is not correct - required format: \
+                    step$typeOfDistribution$numberOfSteps$ci95value$ci95$support, \
```
fso42, line 373: Not sure what the additional step parameter at the beginning is for? This is nowhere mentioned, at least I cannot find it?!

fso42 (changes requested):
Please also add  info about setup to the documentation

avaframe/com1DFA/deriveParameterSet.py
awirb: this is needed to decide which step of the distribution is chosen, as the distribution is only build once we are already in the simDict for each single sim - this step is appended inside the function that creates the configuration entry for this, and on the other hand it can be also provided right from the start if only one step is desired

awirb: splitVariationToArraySteps there it is appended if not provided already but it is right in the error message I removed the step - because it is actually in the if for not providing the step so removed the step there

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -346,6 +358,24 @@ def splitVariationToArraySteps(value, key, fullCfg):
             itemsArray = np.linspace(0.0, float(itemsL[0]), int(itemsL[1]))
         else:
             itemsArray = np.linspace(-1.*float(itemsL[0]), float(itemsL[0]), int(itemsL[1]))
+    # if variaiton following normal distribution
+    elif 'Dist' in key:
+        itemsArray = []
+        # if not already appended the step of the distribution values that shall be taken as value
```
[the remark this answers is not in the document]
awirb: updated the docstring

avaframe/in1Data/computeFromDistribution.py
```diff
@@ -120,3 +123,71 @@ def getEmpiricalCDFNEW(sample):
...
+    # if cfg['flagMinMax'] == 'False':
+    #     sampleVect = sampleVect[1:-1]
+
+    # Debug plot
+    # plt.title('Samples drawn from dist')
+    # plt.plot(sampleVect, 'k*')
+    # plt.xlabel('number of sample')
+    # plt.ylabel('relTh [m]')
+    # plt.show()
+
```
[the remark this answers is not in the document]
awirb: I removed the debug plot, do we want to keep the flagMinMax here?

avaframe/in1Data/computeFromDistributionCfg.ini
```diff
@@ -8,14 +8,24 @@ a = 0.0
...
+# name of distribution i.e. pert [Pert], uniform [Uni], normal [Norm]
 distType = Pert
+
+# NORMAL DIST parameters
+# mean
+mean = 50.
+# needs to be based on mean and std OR 95% confidence interval defined by buildType: std, ci95
+buildType = ci95
+# value of buildType (std or ci95)
+buildValue = 0.3
```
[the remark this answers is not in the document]
awirb: do you have a suggestion? it should be the value for the buildType that is set in the previous line so either std or ci95 to build the distribution

avaframe/com1DFA/deriveParameterSet.py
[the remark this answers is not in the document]
awirb: and made a separate function for addStandardConfig, also included a warning and info to the cfg files that the parameter addStandardConfig is only availabe for percent or range var.

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -444,6 +517,19 @@ def setVariationForAllFeatures(cfg, key, thType, varType, variationFactor):
     idList = cfg['INPUT'][thType + 'Id'].split('|')
     # fetch thickness list
     thicknessList = cfg['INPUT'][thType + 'Thickness'].split('|')
+    # fetch ci95 list
+    ci95List = cfg['INPUT'][thType + 'Ci95'].split('|')
+
+    # do some preprocessing if varTypu is dist
```
fso42, line 523:
```suggestion
    # do some preprocessing if varType is dist
```

avaframe/in1Data/computeFromDistribution.py
```diff
@@ -7,6 +7,8 @@
 import scipy.special as sc
 from scipy.interpolate import interp1d
 import logging
+from scipy.stats import norm
+import matplotlib.pyplot as plt
```
fso42, line 11: Please remove this import. It is not needed, right?

avaframe/in1Data/computeFromDistribution.py
```diff
@@ -120,3 +123,64 @@ def getEmpiricalCDFNEW(sample):
...
+    CDFint = interp1d(cdf, x)
+    # draw sampleSize samples from distribution
+    # also consider if min and max value shall be part of the final sample
+    # TODO: do we need this flagMinMax?
+    # if cfg['flagMinMax'] == 'False':
+    #     sampleSize = sampleSize + 2
+    xSample = np.linspace(np.min(cdf), np.max(cdf), sampleSize)
+    sampleVect = CDFint(xSample)
+    # if cfg['flagMinMax'] == 'False':
+    #     sampleVect = sampleVect[1:-1]
```
fso42, line 184: Is this comment still needed?

avaframe/in1Data/computeFromDistributionCfg.ini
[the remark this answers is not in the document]
fso42: Maybe write an example in the comment?

fso42 (changes requested):

avaframe/runScripts/runProbAnaCom1DFA.py
```diff
@@ -79,7 +79,7 @@
         log.warning('No files found for configuration: %s' % probConf)
```
matthiasto, line 79: line 65, you also remove the work directory, is this wanted?

avaframe/ana3AIMEC/aimecTools.py
```diff
@@ -121,6 +121,8 @@ def fetchReferenceSimNo(avaDir, inputsDF, comModule, cfgSetup):
         colorVariation: boolean
             True if a color variation should be applied in the plots
     """
+
+    cfgSetup = cfg['AIMECSETUP']
```
matthiasto, line 125: is it a good idea to give the full cfg instead of the cfgSetup? an option could be to pass the cfgSetup in input and an optional cfg['filter]

avaframe/ana3AIMEC/ana3AIMECCfg.ini
```diff
@@ -65,6 +65,13 @@ referenceSimName =
 unit =
 
 #---------------------------------------
+## Uncomment this section FILTER in your local copy of the ini file and add filter parameter and parameter values
+## see the example provided below for release thickness
+#[FILTER]
+## define parameter and corresponding values from the simulation configuration to filter simulations
+## multiple parameters are possible, just add them in a new line each
+##relTh = 0.75|0.9|0.8
```
matthiasto, line 73: if I uncomment this, it will filter me all simulations with relTH= to 0.75 or 0.9 or 0.8?

matthiasto (commented):
I tried the probaCom1DFA, seems to run.
I still do not understand why I get 4 output folders: ana4Stats ana4Stats_testall ana4Stats_testMu and ana4Stats_testRelTH.

avaframe/in1Data/computeFromDistributionCfg.ini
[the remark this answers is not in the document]
awirb: I added an example in the comment: 

##### needs to be based on mean and std OR 95% confidence interval defined by buildType: std, ci95
buildType = ci95
###### value of buildType (std or ci95) e.g. if buildType=std and the standard deviation is 0.3,  buildValue=0.3
buildValue = 0.3

avaframe/ana4Stats/probAna.py
```diff
@@ -96,37 +103,74 @@ def updateCfgRange(cfg, cfgProb, varName):
...
             cfg['GENERAL'][varName] = '%s$%s$%s' % (valVal, valVariation, valSteps)
             valValues = fU.splitIniValueToArraySteps(cfg['GENERAL'][varName])
-        else:
-            valStart = str(float(valVal) - float(valVariation))
-            valStop = str(float(valVal) + float(valVariation))
+
+        elif cfgProb['PROBRUN']['variationType'].lower() == 'range':
+            if '-' in valVariation or '+' in valVariation:
+                valStart = str(float(valVal) + float(valVariation))
+                valStop = float(valVal)
```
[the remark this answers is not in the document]
awirb: yes I think so, I needed it to increase the number of particles, which I only wanted to increase not decrease, it also happens for the TH values but at a later point due to the functionality needed of being able to readFromShp

avaframe/ana4Stats/probAnaCfg.ini
```diff
@@ -22,23 +22,40 @@ peakLim = 1.0
...
+flagMinMax = True
+# steps that are used to derive the distribution, should be quite high
+support = 10000
+# NORMAL DIST parameters
+# normal distribution build from std or confidence interval 95%
+buildType = ci95
+# min and max interval [%] for distribution to draw samples from
+minMaxInterval = 95
+
+
```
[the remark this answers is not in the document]
awirb: good point, I put them here to make it more visible and easier to change at a glance, if we use the computeFromDistribution config, one needs to be aware of this and that there is nothing special in the local and if we just read the default, then the parameters can't be changed.. so I thought this is the best way to go?

avaframe/in1Data/computeFromDistribution.py
awirb: removed

awirb:
probabilityConfigurations = {'testAll': {}, 'testRelTh': {'mu': '0.15500',
    'simTypeActual': 'null'}, 'testMu': {'relTh': '1.', 'simTypeActual': 'null'}}

because of this it creates the different output folders, so you get a  probability map for all results, for the ones where relTh is varied, for the ones where mu is varied

avaframe/runScripts/runProbAnaCom1DFA.py
awirb: yes, because otherwise it throws an error if a workDir is already existing, as it first calls initialiseDirs in com1DFAmain:
 in initialiseDirs:
if workDir.is_dir():
        message = 'Work directory %s already exists - delete first!' % (workDir)
        log.error(message)
        raise AssertionError(message)

avaframe/ana3AIMEC/aimecTools.py
awirb: why is this better?, I thought adding a second input parameter to the function is not needed if we just pass the full cfg

avaframe/ana3AIMEC/ana3AIMECCfg.ini
awirb: yes

matthiasto (approved):

avaframe/ana4Stats/probAnaCfg.ini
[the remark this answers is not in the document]
fso42: In my opinion, the override should only be applied for stuff that HAS to be in a certain way for the overriding module to work properly. Which is not the case here, right?

avaframe/in1Data/computeFromDistribution.py
awirb: removed

avaframe/ana4Stats/probAnaCfg.ini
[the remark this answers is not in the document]
awirb: right, I also didn't apply the override correctly be loading the moduleConfig and applying the override there, changed this now and added only those parameter that have to be there to work correctly as so far we only allow the thickness values to have the attribute ci95 - hence we need this as buildType

fso42 (approved):

PR #754, opened 2022-08-23
Prepare com1DFA code according to theory paper

matthiasto:
This PR will update the code so that it matches the description in the theory paper.
I added com1DFA options:
- to add the curvature in the pressure gradient computation
- to add the curvature in the friction force
- to add the curvature in the tangential equation

The default settings should still produce the benchmarks results


fso42 (changes requested):

fso42:
To keep track: commit https://github.com/avaframe/AvaFrame/commit/419c11fb3f67e3730eee074319af42c3f3e13067 introduced some changes to the benchmark results in regards to remeshing -> separate issue!

fso42 (approved):

PR #757, opened 2022-08-29
add scenario name to configuration that can be used for plotting example

awirb:


fso42:
@awirb: Why not just use the simulation name?

matthiasto (commented):
What does this scenario name say? can you give an example?

awirb:
It can be used to gather multiple simulations for example if you do a probRun, and vary multiple variables one after the other, and you want to plot some result parameter vs the other and group the ones where parameter a was varied and those where parameter b was varied.

fso42 (approved):
Maybe squash the commits to just one...

PR #775, opened 2022-10-13
Adress filter/aimec bugs and add resDF

awirb:
* fix some bug for filtering function 
* add comments to stats plots 
* fix bug in aimec warning
* add resDF to be saved for com1DFA results that can be used for statistical analysis

### DIFF
--- a/avaframe/com1DFA/com1DFA.py
+++ b/avaframe/com1DFA/com1DFA.py
@@ -1311,6 +1311,9 @@ def DFAIterate(cfg, particles, fields, dem, simHash=''):
     massEntrained = []
     massTotal = []
 
+    # setup a result fields info data frame to save max values of fields and avalanche front
+    resultsDF = setupresultsDF(resTypesLast, cfg['VISUALISATION'].getboolean('createRangeTimeDiagram'))
+
     # TODO: add here different time stepping options
     log.debug('Use standard time stepping')
     # Initialize time and counters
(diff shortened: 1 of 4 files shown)


avaframe/com1DFA/com1DFA.py
```diff
@@ -1311,6 +1311,9 @@ def DFAIterate(cfg, particles, fields, dem, simHash=''):
     massEntrained = []
     massTotal = []
 
+    # setup a result fields info data frame to save max values of fields and avalanche front
+    resDF = setupResDF(resTypesLast, cfg['VISUALISATION'].getboolean('createRangeTimeDiagram'))
```
fso42, line 1315: Would you mind renaming this to something like resultsDF resuDF ... (same for the functions)?
res is very much preloaded with 'resistance in the context of com1dfa...

avaframe/com1DFA/com1DFA.py
```diff
@@ -1447,9 +1456,76 @@ def DFAIterate(cfg, particles, fields, dem, simHash=''):
...
+            resDict['max' + resT] = [0.0]
+    if cfgRangeTime:
+        resDict['rangeList'] = [0.0]
+    resDF = pd.DataFrame.from_dict(resDict)
+    resDF = resDF.set_index('timeStep')
+
+    return resDF
+
+
+def getMaxMeanValues(resDF, fields, timeStep, resTypes, rangeValue=''):
```
fso42, line 1495: We already have a getMaxMeanValues function in aimecTools, could that be used?

fso42 (changes requested):

avaframe/com1DFA/com1DFA.py
awirb: Actually the getMaxMeanValues function here just appends the max value of the fields to the dataFrame and if provided the rangeValue, the actual computation of the max value is just one line and so I think not useful to call the aimec function, however, the function here could be renamed to addMaxValuesToDF?

avaframe/com1DFA/com1DFA.py
awirb: renamed to resultsDF and setupResultSDF now

fso42 (approved):

PR #801, opened 2023-01-05
fix unexpected keyword error

ninsbl:
With Python 3.8.10 (default, Nov 14 2022, 12:59:47) and Matlplotlib 3.3.4 I get an "unexpected keyword error" The brackets should fix issue. See also:
https://stackoverflow.com/questions/37280557/can-i-place-a-vertical-colorbar-to-the-left-of-the-plot-in-matplotlib#comment112236764_61090677


fso42:
Thanks for your contribution @ninsbl!

PR #826, opened 2023-03-06
Move check for default indicator [com1DFA]; fixes #825

### LINKED ISSUE #825: MeshCellSize leads to all sims having the _C_ changed indicator
Logic error, since this is actually the default:
{"root['GENERAL']['sphKernelRadius']": {'new_value': '5', 'old_value': 'meshCellSize'}}
7413, 139670472533824 - com1DFATools - INFO - Modified from default

Expected behaviour: setting the meshCellSize should be recognized as default...

fso42:


fso42:
Standardtest ok

PR #829, opened 2023-03-08
Prepare ana4ProbAna for QGis connector

fso42:
- Prepare ana4ProbAna for connector usage
- renames outputs and locations of plots and results
- drop quite a few logs from input to debug
DONE: @awirb : please have a look at function fetchProbConfigs: I think there is an error with the section of the cfg used... I think the cfg['varParList'] needs to bee cfg['PROBRUN']['varParList'] or similar


avaframe/com1DFA/com1DFA.py
```diff
@@ -361,7 +361,7 @@ def com1DFACore(cfg, avaDir, cuSimName, inputSimFiles, outDir, simHash=''):
     # set thickness values for the release area, entrainment and secondary release areas
     relName, inputSimLines, badName = prepareReleaseEntrainment(cfg, inputSimFiles['releaseScenario'], inputSimLines)
 
-    log.info('Perform %s simulation' % cuSimName)
+    log.debug('Perform %s simulation' % cuSimName)
```
awirb, line 364: I think this is a really helpful comment in the log and while running the sim - an option to keep it on info level?

avaframe/com1DFA/particleTools.py
```diff
@@ -214,7 +214,7 @@ def removePart(particles, mask, nRemove, reasonString='', glideSnow=0):
         particles dictionary
     """
     if reasonString != '':
-        log.info('removed %s particles %s' % (nRemove, reasonString))
+        log.debug('removed %s particles %s' % (nRemove, reasonString))
```
awirb, line 217: I think this is also an important message, as sometimes there is a problem and particles are then also removed so it is a good hint that something could be wrong!

avaframe/out3Plot/statsPlots.py
```diff
@@ -298,7 +300,12 @@ def plotProbMap(avaDir, inDir, cfgFull, demPlot=False):
 
         # create figure and add title
         fig = plt.figure(figsize=(pU.figW*2, pU.figH))
-        fullTitle = cfg['name'] + ' based on %s $>$ %s %s' % (cfgFull['GENERAL']['peakVar'], cfgFull['GENERAL']['peakLim'], cfgFull['GENERAL']['unit'])
+
+        fullTitle = '%s %s based on %s $>$ %s %s' % (avaName,
```
awirb, line 304: or shall we add the avaName on the plot (inside) instead of making the title longer?

avaframe/runAna4ProbAna.py
```diff
@@ -38,27 +39,34 @@ def runProbAna(avalancheDir=''):
     # Load general configuration filee
     cfgMain = cfgUtils.getGeneralConfig()
 
-    # Read avalanche directory
-    avaDir = cfgMain['MAIN']['avalancheDir']
-    avaDir = pathlib.Path(avaDir)
-    avaName = avaDir.name
+    # Load avalanche directory from general configuration file
+    # More information about the configuration can be found here
```
awirb, line 43: here? did you want to put the link?

avaframe/runAna4ProbAna.py
```diff
@@ -38,27 +39,34 @@ def runProbAna(avalancheDir=''):
     # Load general configuration filee
     cfgMain = cfgUtils.getGeneralConfig()
 
-    # Read avalanche directory
-    avaDir = cfgMain['MAIN']['avalancheDir']
-    avaDir = pathlib.Path(avaDir)
-    avaName = avaDir.name
+    # Load avalanche directory from general configuration file
```
awirb, line 42: suggestion add: "if not provided as input"

avaframe/runAna4ProbAna.py
```diff
@@ -38,27 +39,34 @@ def runProbAna(avalancheDir=''):
...
-    initProj.cleanSingleAvaDir(avaDir, keep=logName, deleteOutput=False)
+    initProj.cleanSingleAvaDir(avalancheDir, keep=logName, deleteOutput=False)
 
     # Load configuration file for probabilistic run and analysis
     cfgProb = cfgUtils.getModuleConfig(probAna)
 
-    # create configuration files for com1DFA simulations including parameter variation - defined in the probabilistic config
-    cfgFiles, cfgPath = probAna.createComModConfig(cfgProb, avaDir, com1DFA, cfgFileMod='')
+    # create configuration files for com1DFA simulations including parameter
+    # variation - defined in the probabilistic config
```
awirb, line 65: shall we be more specific and say in probAnaCfg.ini? or its local copy?

avaframe/tests/test_probAna.py
```diff
@@ -47,10 +47,10 @@ def test_probAna(tmp_path):
 
     # call function to test
     pA.probAnalysis(avaDirtmp, cfg, com1DFA, parametersDict=parametersDict, inputDir='')
-    probTest = np.loadtxt(os.path.join(avaDirtmp, 'Outputs', 'ana4Stats', 'avaParabola_probMap1.0.asc'), skiprows=6)
+    probTest = np.loadtxt(os.path.join(avaDirtmp, 'Outputs', 'ana4Stats', 'avaParabola_prob__ppr_lim1.0.asc'), skiprows=6)
```
awirb, line 50: wouldn't this now need to include the probConf?

awirb (changes requested):

awirb:
regarding the fetchProbConf() function: in the runAna4ProbAna.py - it should be cfgProb['PROBRUN'] that is passed to the function.. and as in the other scripts the probConfs are renamed to includeXX instead of testXX, also in this function the test could be renamed to include

avaframe/com1DFA/com1DFA.py
fso42: For most applications, this is already printed via a log.info on com1DFACoreTask, which has, IMHO higher priority due to the parallel execution. If someone calls core directly, I think it is fair to say that they can drop to debug in case of problems...

avaframe/com1DFA/particleTools.py
fso42: Then we need to collect this information and make a summary log AFTER the simulation is done. Printing it directly during execution significantly slows down execution since it creates a bottleneck on the standard output, especially for parallel runs. I'll start an issue.

avaframe/out3Plot/statsPlots.py
fso42: Mh, for the cases I had it worked out fine, but we can also move it to the inside of the plot.

avaframe/tests/test_probAna.py
fso42: If we'd provide it in the test, then yes, but right now it is not included...

fso42:
> regarding the fetchProbConf() function: in the runAna4ProbAna.py - it should be cfgProb['PROBRUN'] that is passed to the function.. and as in the other scripts the probConfs are renamed to includeXX instead of testXX, also in this function the test could be renamed to include

I'm not sure what you mean by this? Where should testXX be renamend to includeXX?

awirb (approved):

PR #839, opened 2023-03-23
add optional argument for general config

awirb:


fso42 (approved):

fso42 (changes requested):
Mh, this is a pretty central function,  I think it should be tested in the pytests...

fso42 (approved):

PR #841, opened 2023-04-11
split and add line point snapping distance functions

awirb:


avaframe/in3Utils/geoTrans.py
```diff
@@ -547,6 +541,75 @@ def findSplitPoint(avaProfile, Points):
     return projPoint
 
 
+def findClosestPoint(xcoor, ycoor, pointsDict):
+    """ find closest point of pointDict along line defined by xcoor and ycoor - only xy plane!
```
fso42, line 545: Maybe add info that pointsDict  has to contain 'x' and 'y'

fso42 (approved):

avaframe/in3Utils/geoTrans.py
awirb: added

Issue #586, opened 2021-11-11
Add tool to export Release Area data (for thickness evaluation)


fso42:
- [ ] Maybe as qgis script as well?
- [ ] include whole standard 
- [ ] MSC (maybe own issue)

fso42:
[Schneehöhenermittlung.txt](https://github.com/avaframe/AvaFrame/files/11321455/Schneehohenermittlung.txt)
This is an example for the resulting textfile/info

PR #845, opened 2023-04-27
add creating rel info file [com1DFA]

awirb:
address issue #586

### DIFF
--- a/avaframe/in1Data/getInput.py
+++ b/avaframe/in1Data/getInput.py
@@ -7,12 +7,18 @@ import os
 import glob
 import pathlib
 import logging
+import numpy as np
+import pandas as pd
+import shapely as shp
 
 # Local imports
 from avaframe.in3Utils import cfgUtils
 import avaframe.in2Trans.ascUtils as IOf
 import avaframe.in2Trans.shpConversion as shpConv
 import avaframe.com1DFA.deriveParameterSet as dP
+import avaframe.com1DFA.DFAtools as DFAtls
+import avaframe.in3Utils.fileHandlerUtils as fU
+from avaframe.com1DFA import com1DFA
 
 
 # create local logger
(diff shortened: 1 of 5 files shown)


avaframe/in1Data/getInput.py
```diff
@@ -7,12 +7,19 @@
 import glob
 import pathlib
 import logging
+import numpy as np
+import matplotlib.pyplot as plt
```
fso42, line 11: Unused, please remove

avaframe/in1Data/getInput.py
```diff
@@ -497,3 +504,169 @@ def fetchReleaseFile(inputSimFiles, releaseScenario, cfgSim, releaseList):
...
+    header = dem['header']
+    csz = header['cellsize']
+
+    # define parameters to be saved
+    relFeatures = []
+    relMaxZ = []
+    relMinZ = []
+    relSlope = []
+    relArea = []
+    relAreaP = []
```
fso42, line 544: To make this more readable, I suggest switching to a dict with keys here, especially in connection with my comment below

avaframe/in1Data/getInput.py
```diff
@@ -497,3 +504,169 @@ def fetchReleaseFile(inputSimFiles, releaseScenario, cfgSim, releaseList):
...
+
+            # print info
+            log.info('++++ Release feature %s ++++++++++' % releaseLine['Name'][index])
+            log.info('release area projected: %.2f' % areaProjectedList[index])
+            log.info('release area projected using shapely: %.2f' % projectedAreas[index])
+            log.info('release area real: %.2f' % areaActualList[index])
+            log.info('maximum elevation: %.2f' % zMax)
+            log.info('mean elevation: %.2f' % zMean)
+            log.info('minimum elevation: %.2f' % zMin)
+            log.info('mean slope: %.2f' % meanSlope)
```
fso42, line 594: I think this should be packed into a separate function `computeRelElevStats` (or something like this) which returns a dict with keys, which is then use to fill relD

fso42, line 594: To keep the clatter in the log to a minimum, I think this should be dropped to debug. Especially since this information is saved and easily readable info about it is given in the log.info below...

avaframe/runScripts/runReleaseInfo.py
```diff
@@ -0,0 +1,42 @@
+"""
+    Run script for creating a release area info file 
+"""
+
+# Local imports
+import avaframe.in3Utils.initializeProject as initProj
```
fso42, line 6: unused, remove

avaframe/runScripts/runReleaseInfo.py
```diff
@@ -0,0 +1,42 @@
...
+from avaframe.com1DFA import com1DFA
+import avaframe.com1DFA.deriveParameterSet as dP
```
fso42, line 8: unused, remove

avaframe/runScripts/runReleaseInfo.py
```diff
@@ -0,0 +1,42 @@
...
+from avaframe.in3Utils import cfgUtils
+from avaframe.in3Utils import logUtils
+import avaframe.in1Data.computeFromDistribution as cF
+import avaframe.out3Plot.in1DataPlots as iP
+import avaframe.out3Plot.statsPlots as sPlot
```
fso42, line 13: unused, remove

avaframe/runScripts/runReleaseInfo.py
```diff
@@ -0,0 +1,42 @@
...
+import avaframe.in1Data.getInput as gI
+
+import numpy as np
+from scipy.stats import norm
+import configparser
+import matplotlib.pyplot as plt
+import pandas as pd
+import pathlib
+import matplotlib.pyplot as plt
```
fso42, line 22: unused, remove

fso42 (changes requested):

avaframe/in1Data/getInput.py
awirb: moved to a separate func

awirb: I think it is the only output of the runScript.. but for example all the parameters of the com1DFA config are listed in the log although only the methodMeshNormal is used - maybe we could decrease this info in the log, but I think it is quite nice to see the stats printed and not have to open the file - but I moved it to debug

fso42 (approved):

PR #862, opened 2023-06-13
remove unused filter option, delete output for ana4Prob run script

awirb:
-> to discuss should deleteOutput=True for runAna4ProbAna.py - will remove Outputs in avaDir but if not also old com1DFA sims will be used when computing the probMaps.. which might not be intended

* remove Filter option from probAnaCfg.ini as it currently gets overwritten in the runAna4ProbAna.py runcscript which is useful for the one at a time variation option 
-> if desired to keep filter option - also possible by adding instead of overwriting


fso42:
@awirb : could we make use of the latestSims.csv (https://github.com/avaframe/AvaFrame/blob/1860cd98d994e0fb4dc05904a2c017a50ecdb15f/avaframe/com1DFA/com1DFA.py#L302) instead of deleting the outputs folder?

avaframe/runAna4ProbAna.py
```diff
@@ -55,7 +55,7 @@ def runProbAna(avalancheDir=''):
     log.info('Current avalanche: %s', avalancheDir)
 
     # Clean input directory(ies) of old work files
-    initProj.cleanSingleAvaDir(avalancheDir, keep=logName, deleteOutput=False)
+    initProj.cleanSingleAvaDir(avalancheDir, keep=logName, deleteOutput=True)
```
fso42, line 58: Please see my general comment

fso42 (changes requested):

fso42 (approved):

fso42:
Standardtest ok

PR #886, opened 2023-07-13
Prob plots [com1; out3; ana4]

awirb:
* update runAna4ProbAna – com1DFA settings from com1DFA_override in probAnaCfg.ini and either default com1DFA cfg if com1DFA_override – defaultConfig=True if not and a local_com1DFACfg is available use this too combine with parameters in com1DFA_override

*  check if desired simType for com1DFA is a valid option [null, ent, res, entres, available]

* contour line plots – if one contour line is made up by multiple lines – make it still a nice plot

* create a sample plot for probAna if the created sample has a variation of two parameters
but one of them is a thickness parameter and has multiple featues e.g. relTh0, relTh1, 

the thickness is varied for the individual thickness features in the same way (same percent variation) and add now a plot that has multiple y axis to cover all the features


avaframe/out3Plot/outAIMEC.py
```diff
@@ -992,20 +992,48 @@ def plotContoursTransformed(contourDict, pathDict, rasterTransfo, cfgSetup):
             cmapVal = simDF.loc[simDF['simName'] == simName, paraVar]
         else:
             cmapVal = index / len(contourDict)
-        if simName == pathDict['refSimName']:
-            # define label for reference sim, if info on what it is based is available
-            if pathDict['valRef'] == '':
-                labelReference = 'reference'
+        for key in contourDict[simName]:
+            print('key', key, 'simName', simName)
```
fso42, line 996: seems to be a debug print? remove?

PR #905, opened 2023-09-08
allow to pass dem [ana3]

awirb:
* pass path to dem
* if two sims to compare but not all three result variables provided still create plot


avaframe/ana3AIMEC/ana3AIMEC.py
```diff
@@ -20,7 +20,7 @@
 log = logging.getLogger(__name__)
 
 
-def fullAimecAnalysis(avalancheDir, cfg, inputDir=''):
+def fullAimecAnalysis(avalancheDir, cfg, inputDir='', demFileName=''):
```
fso42, line 23: Please add the demFileName to the docstring

fso42 (changes requested):

PR #924, opened 2023-10-08
Adjust float format of ReleaseStat to avoid Excel Problem

fso42:
Having only 2 decimals leads to Excel changing these fields to dates.... I know...

- to avoid any confusion, it is set to 4 decimals now
- clean getInput.py with black and remove unused parameters

### DIFF
--- a/avaframe/in1Data/getInput.py
+++ b/avaframe/in1Data/getInput.py
@@ -27,7 +27,7 @@ log = logging.getLogger(__name__)
 
 
 def readDEM(avaDir):
-    """ read the ascii DEM file from a provided avalanche directory
+    """read the ascii DEM file from a provided avalanche directory
 
     Parameters
     ----------
@@ -43,15 +43,15 @@ def readDEM(avaDir):
     # get dem file name
     demSource = getDEMPath(avaDir)
 
-    log.debug('Read DEM: %s' % demSource)
+    log.debug("Read DEM: %s" % demSource)
 
     dem = IOf.readRaster(demSource)
 
-    return(dem)
+    return dem
 
 
 def getDEMPath(avaDir):
-    """ get the DEM file path from a provided avalanche directory
+    """get the DEM file path from a provided avalanche directory
 
     Parameters
     ----------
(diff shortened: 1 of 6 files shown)


avaframe/in1Data/getInput.py
```diff
@@ -65,59 +65,59 @@ def getDEMPath(avaDir):
...
-        demFile: pathlib path
-            path to dem file
+    Returns
+    --------
+    demFile: pathlib path
+        path to dem file
     """
 
-    if fileName == '':
-        demFilePath = getDEMPath(avaDir)
```
awirb, line 107: isn't this option lost in the new code? that if no fileName provided it looks for the dem in Inputs?

avaframe/in1Data/getInput.py
```diff
@@ -145,59 +145,57 @@ def getInputData(avaDir, cfg):
...
                 raise FileNotFoundError(message)
             relFiles.append(relf)
-        log.debug('Release area file is specified to be: %s' % relFiles)
+        log.debug("Release area file is specified to be: %s" % relFiles)
     else:
-        releaseDir = 'REL'
-        relFiles = sorted(glob.glob(inputDir+os.sep + releaseDir+os.sep + '*.shp'))
-    log.info('Release area files are: %s' % relFiles)
+        releaseDir = "REL"
+        relFiles = sorted(glob.glob(inputDir + os.sep + releaseDir + os.sep + "*.shp"))
```
awirb, line 171: switch to pathlib instead of os?

avaframe/in1Data/getInput.py
```diff
@@ -516,69 +509,65 @@ def fetchReleaseFile(inputSimFiles, releaseScenario, cfgSim, releaseList):
...
     # loop over all relFiles and compute information saved to dictionary of dataframes
-    relInfo = {}
-    relPath = pathlib.Path(avaDir, 'Outputs', 'com1DFA', 'releaseInfoFiles')
+    relPath = pathlib.Path(avaDir, "Outputs", "com1DFA", "releaseInfoFiles")
     fU.makeADir(relPath)
     relDFDict = {}
     # loop over all relFiles
     for relFile in relFiles:
         # for relF in relFiles:
-        releaseLine = {}
```
awirb, line 576: if it is not reinitialized as empty dict, doesnt it add/overwrite the keys?

awirb (changes requested):

avaframe/in1Data/getInput.py
fso42: demFilePath is never used, so this if makes no sense at all and only leads to an empty / undefined demFile being returned. Or am I missing something here?

avaframe/in1Data/getInput.py
fso42: Mh, I'm pretty sure the keys are being overwritten, which is exactly what we want, right?

avaframe/in1Data/getInput.py
fso42: ok, I'll include this then and make a pytest for it. Since we obviously missed it so far :-)

PR #934, opened 2023-10-24
update docu

awirb:


docs/moduleCom1DFA.rst
```diff
@@ -106,6 +107,11 @@ Release, entrainment and secondary release thickness can be specified in two dif
     ci95). In order to use this variation, set the 'THICKESSRangeFromCiVariation'
     to ``ci95$numberOfSteps``.
 
+.. Note::
+   In case of the entrainment area, if NO thickness attribute is available
+   the entrainment thickness is set to the value of `entThIfMissingInShp` (default 0.3 m) in the `com1DFACfg.ini` file
+   (or your local copy).
+
```
fso42, line 114: Please remove this, I think the entry in the list above is enough...

fso42 (changes requested):

docs/moduleCom1DFA.rst
awirb: forgot to remove this after the change, removed now

PR #950, opened 2023-12-13
change finding of velocity threshold indices in aimec for thalweg plots [ana3]

awirb:
change how indVelZero index is found for computation of runoutAngle in thalweg plot (just for plot) angle in dataframe is computed differently as it is based on runoutResType and not on velocity but now same method i used

### DIFF
--- a/avaframe/ana3AIMEC/ana3AIMECCfg.ini
+++ b/avaframe/ana3AIMEC/ana3AIMECCfg.ini
@@ -24,6 +24,9 @@ resTypes = ppr|pft|pfv
 # data result type for runout analysis (ppr, pft, pfv)
 runoutResType = ppr
 
+# show cross profile max or mean values on slComparisonstat plot (options: Max, Mean)
+runoutCrossType = Max
+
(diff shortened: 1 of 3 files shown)


avaframe/ana3AIMEC/ana3AIMECCfg.ini
```diff
@@ -24,6 +24,9 @@ resTypes = ppr|pft|pfv
 # data result type for runout analysis (ppr, pft, pfv)
 runoutResType = ppr
 
+# show cross profile max or mean values on slComparisonstat plot
```
fso42, line 27: please indicate what the options are, i.e. how the parameter needs to be set to e.g. get the mean values...

fso42 (changes requested):

avaframe/ana3AIMEC/ana3AIMECCfg.ini
awirb: done

fso42 (approved):

PR #952, opened 2024-01-04
allow for multiple relTH files and search for matching one

awirb:
* allow multiple RELTH files with different spatial resolution in Inputs/RELTH folder
* when creating simDicts, set path to matching relTh file in configuration 
* this is then loaded when the inputs are loaded for each sim


fso42:
Included changes in #954 ; closing this one

PR #959, opened 2024-01-31
add convert to th from depth [in2; out3]

awirb:


avaframe/in2Trans/transfromFields.py
```diff
@@ -0,0 +1,112 @@
...
+# local imports
+import avaframe.com1DFA.DFAtools as DFAtls
+from avaframe.in3Utils import geoTrans as gT
+import avaframe.in2Trans.ascUtils as IOf
+
+# create local logger
+log = logging.getLogger(__name__)
+
+
+def convertDepthToThickness(depthDict, demDict, num=6):
```
fso42, line 19: Mh, can we somehow get rid of the num parameter? Especially since it is different than the default from the com1 ini?
Ideas: set this to a default in the original function, or read the ini, or...?

avaframe/in2Trans/transfromFields.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+
+    if not pathlib.Path(filePath).is_file():
+        message = "Provided filePath %s is not a file" % str(filePath)
+        log.error(message)
+        raise FileNotFoundError(message)
+    else:
+        filePath = pathlib.Path(filePath)
+
+    return filePath
```
fso42, line 78: What is this needed for?

avaframe/in2Trans/transfromFields.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def fetchPointValuesFromField(dataDF, xyPoints, resType, interpMethod='bilinear'):
```
fso42, line 81: The interpMethod does not do anything. Either remove it or hand it over to the project function

avaframe/runScripts/runD2Th.py
```diff
@@ -0,0 +1,63 @@
...
+    for rF in resFiles:
+        # read depth field to dict
+        depthField = IOf.readRaster(rF)
+
+        # convert depth to thickness using dem
+        thicknessDict, depthRasterResized, slopeAngleField = tF.convertDepthToThickness(depthField, dem, num=8)
+
+        # create output directory
+        outDir = pathlib.Path(avaDir, "Outputs", comMod, "peakFiles", "transformed")
+        fU.makeADir(outDir)
```
fso42, line 34: This could go outside the loop, right?

fso42 (changes requested):

avaframe/in2Trans/transfromFields.py
awirb: it was first included in a function now not anymore but I think this can now be used whenever we check if a file exists and want to throw an error if not

avaframe/in2Trans/transfromFields.py
awirb: this was a bug interp in line 107 is now set to interpMethod

avaframe/in2Trans/transfromFields.py
fso42: I'm pretty sure there's better function in pathlib for that?

fso42 (approved):

fso42 (changes requested):
Shall we add the illustration from the com1DFACfg.ini regarding the 'num' parameter to the function?

PR #966, opened 2024-04-02
Add +1SD point in com2AB plots [com2AB]

fso42:


awirb (approved):

fso42:
Standardtests ok, apart from the KOT differences as before

PR #971, opened 2024-05-10
adjust format of saving particle csv files

awirb:
+ fix issue #971


fso42:
Could you address issue #890 in here as well? Would fit thematically....

fso42 (changes requested):
Mh, I guess we need to rename all benchmark pickle files as well...

fso42:
Standardtests ok, apart from avaKot (known)

PR #977, opened 2024-05-17
Include py 3.12; add psutil requirement

fso42:
Newest QGis uses python 3.12, this updates the release scripts to include 3.12

Thanks @ Sebastian D. for reporting.


fso42:
Standardtests ok (apart from known Kot difference)

PR #990, opened 2024-06-21
[com4FlowPy]: add travel length to output-files

PaulaSp3:
store travel length of every path and save it as output (as CH did in the old repo)


fso42 (approved):

ahuber-bfw:
Looks good to me!
Just two minor remarks/comments to keep in mind:

- Maybe we can think about having the outputs controllable via the .ini file (e.g. provide a list of outputs) --> **future feature enhancement**
- The travel_length output looks reasonable in general, but also has some weird artefacts in some test cases (see picture below) - but this is probably linked to other issues (cell_counts, flux conservation, ...)

image

ahuber-bfw (approved):

Issue #330, opened 2021-02-25
Remove flag from all the flagSOMETHING


fso42:
Not really needed, as it is redundant when using booleans....

PR #991, opened 2024-06-21
[com1DFA]: add detrainment

PaulaSp3:
- For res-simulation: add option to consider detrainment (instead of resistance or additionally)
- add option to get detrained mass as output (colormap for the report/plot does not fit, for resType = dmDet)
- write detrained mass per time step in mass balance file

- [ ] Docu
- [ ] Add additional particle property to doc 
Note @fso42 : Doc goes into extra PR

### DIFF
--- a/avaframe/com1DFA/DFAfunctionsCython.pyx
+++ b/avaframe/com1DFA/DFAfunctionsCython.pyx
@@ -128,6 +128,7 @@ def computeForceC(cfg, particles, fields, dem, int frictType):
   cdef double[:, :] VZ = fields['Vz']
   cdef double[:, :] entrMassRaster = fields['entrMassRaster']
   cdef double[:, :] entrEnthRaster = fields['entrEnthRaster']
+  cdef double[:, :] detRaster = fields['detRaster']
   cdef double[:, :] cResRaster = fields['cResRaster']
   cdef int[:] indXDEM = particles['indXDEM']
   cdef int[:] indYDEM = particles['indYDEM']
@@ -137,11 +138,13 @@ def computeForceC(cfg, particles, fields, dem, int frictType):
   cdef double[:] forceZ = np.zeros(nPart, dtype=np.float64)
   cdef double[:] forceFrict = np.zeros(nPart, dtype=np.float64)
   cdef double[:] dM = np.zeros(nPart, dtype=np.float64)
+  cdef double[:] dMDet = np.zeros(nPart, dtype=np.float64)
   cdef double[:] gEff = np.zeros(nPart, dtype=np.float64)
   cdef double[:] curvAcc = np.zeros(nPart, dtype=np.float64)
   # declare intermediate step variables
   cdef int indCellX, indCellY
-  cdef double areaPart, areaCell, araEntrPart, cResCell, cResPart, uMag, uMagRes, m, dm, h, entrMassCell, entrEnthCell, dEnergyEntr, dis
+  cdef double areaPart, areaCell, areaEntrPart, cResCell, cResPart, uMag, uMagRes, m, dm, h, entrMassCell, entrEnthCell, dEnergyEntr, dis
+  cdef double dmDet, detCell, areaDetPart
   cdef double x, y, z, xEnd, yEnd, zEnd, ux, uy, uz, uxDir, uyDir, uzDir, totalEnthalpy, enthalpy, dTotalEnthalpy
   cdef double nx, ny, nz, nxEnd, nyEnd, nzEnd, nxAvg, nyAvg, nzAvg
   cdef double gravAccNorm, accNormCurv, effAccNorm, gravAccTangX, gravAccTangY, gravAccTangZ, forceBotTang, sigmaB, tau
@@ -150,6 +153,7 @@ def computeForceC(cfg, particles, fields, dem, int frictType):
   cdef double w[4]
   cdef double wEnd[4]
   cdef int k
+
   force = {}
   # loop on particles
   for k in range(nPart):
@@ -341,6 +345,21 @@ def computeForceC(cfg, particles, fields, dem, int frictType):
           # update specific enthalpy of particle
           totalEnthalpyArray[k] = totalEnthalpy / m
 
+      # adding detrainment analogous to entrainment
+      detCell = detRaster[indCellY, indCellX]
+      
+      if detCell > 0:
+        # compute detrained mass
+        dmDet = computeDetMass(dt, detCell, areaPart, uMag)
+        if abs(dmDet) > m:
+          # mass of a particle can't be negative
+          dmDet = - 1 * m
+      
+        # update mass
+        m = m + dmDet
+        mass[k] = m
+        dMDet[k] = dmDet
+
       # adding resistance force due to obstacles
       cResCell = cResRaster[indCellY][indCellX]
       cResPart = computeResForce(hRes, h, areaPart, rho, cResCell, uMag, explicitFriction)
@@ -352,6 +371,7 @@ def computeForceC(cfg, particles, fields, dem, int frictType):
 
   # save results
   force['dM'] = np.asarray(dM)
+  force['dMDet'] = np.asarray(dMDet)
   force['forceX'] = np.asarray(forceX)
   force['forceY'] = np.asarray(forceY)
   force['forceZ'] = np.asarray(forceZ)
@@ -362,7 +382,9 @@ def computeForceC(cfg, particles, fields, dem, int frictType):
   particles['uy'] = np.asarray(uyArray)
   particles['uz'] = np.asarray(uzArray)
   particles['m'] = np.asarray(mass)
+  particles['dmDet'] = np.asarray(dMDet)
   particles['totalEnthalpy'] = np.asarray(totalEnthalpyArray)
+
 
   # update mass available for entrainement
   # TODO: this allows to entrain more mass then available...
@@ -435,6 +457,40 @@ cpdef (double, double) computeEntMassAndForce(double dt, double entrMassCell,
           areaEntrPart = entrMassCell / rhoEnt
 
   return dm, areaEntrPart
+
+cpdef double computeDetMass(double dt, double detCell,
+                                              double areaPart, double uMag):
+  """ compute detrained mass
+
+  Parameters
+  ----------
+  dt: float
+    time step
+  detCell : float
+      coefficient for detrainment
+  areaPart : float
+      particle area
+  uMag : float
+      particle speed (velocity magnitude)
+
+  Returns
+  -------
+  dm : float
+      detrained mass
+  """
+
+  cdef double dmDet = 0
+  
+  # compute detrained mass
+  if detCell > 0 and uMag > 0:
+      dmDet = - detCell / uMag * dt * areaPart
+  else:
+      dmDet = 0
+
+  if dmDet > 0 or np.isnan(dmDet):
+      log.error('uMag, dt or areaPart is 0')
+
+  return dmDet
 
 
 cpdef double computeResForce(double hRes, double h, double areaPart, double rho,
@@ -633,6 +689,7 @@ def updatePositionC(cfg, particles, dem, force, fields, int typeStop=0):
   cdef double[:] forceSPHY = force['forceSPHY']
   cdef double[:] forceSPHZ = force['forceSPHZ']
   cdef double[:] dM = force['dM']
+  cdef double[:] dMDet = force['dMDet']
   # read dam
   dam = dem['damLine']
   cdef int flagDam = dam['dam']
@@ -672,7 +729,7 @@ def updatePositionC(cfg, particles, dem, force, fields, int typeStop=0):
   cdef double mNew, xNew, yNew, zNew, uxNew, uyNew, uzNew, txWall, tyWall, tzWall, totalEnthalpy, totalEnthalpyNew
   cdef double sCorNew, sNew, lNew, ds, dl, uN, uMag, uMagNew, fNx, fNy, fNz, dv, uMagt0, uMagt1
   cdef double ForceDriveX, ForceDriveY, ForceDriveZ
-  cdef double massEntrained = 0, massFlowing = 0, dissEm = 0
+  cdef double massEntrained = 0, massDetrained = 0, massFlowing = 0, dissEm = 0
   cdef int k, inter
   cdef int nRemove = 0
   # variables for interpolation
@@ -718,6 +775,7 @@ def updatePositionC(cfg, particles, dem, force, fields, int typeStop=0):
     # update mass (already done un computeForceC)
     mNew = m
     massEntrained = massEntrained + dM[k]
+    massDetrained = massDetrained + dMDet[k]
     # update position
     if centeredPosition:
       xNew = x + dtStop * 0.5 * (ux + uxNew)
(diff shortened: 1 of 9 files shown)


avaframe/com1DFA/DFAfunctionsCython.pyx
```diff
@@ -436,6 +458,44 @@ cpdef (double, double) computeEntMassAndForce(double dt, double entrMassCell,
...
+  cdef double width, areaDetPart
+  cdef double dmDet = 0
+  # compute detrained mass
+  if detCell > 0 and uMag > 0:
+      dmDet = - detCell / uMag * dt * areaPart
+  else:
+      dmDet = 0
+
+  if dmDet > 0 or np.isnan(dmDet):
+      dmDet = 0
```
fso42, line 492:
```suggestion

  cdef double width, areaDetPart
  cdef double dmDet = 0
  
  # compute detrained mass
  if detCell > 0 and uMag > 0:
      dmDet = - detCell / uMag * dt * areaPart
  else:
      dmDet = 0

  if dmDet > 0 or np.isnan(dmDet):
      dmDet = 0
```

avaframe/com1DFA/DFAfunctionsCython.pyx
```diff
@@ -888,6 +950,7 @@ def updatePositionC(cfg, particles, dem, force, fields, int typeStop=0):
   particles['y'] = np.asarray(yNewArray)
   particles['z'] = np.asarray(zNewArray)
   particles['massEntrained'] = massEntrained
+  particles['massDetrained'] = massDetrained
```
fso42, line 953: Maybe add a detrained dfa mass debug log as well?

avaframe/com1DFA/DFAfunctionsCython.pyx
```diff
@@ -1112,14 +1179,20 @@ def updateFieldsC(cfg, particles, dem, fields):
       indx = Lx0 + ind1[i]
       indy = Ly0 + ind2[i]
       mwi = m * w[i]
+      dmwi = dm * w[i]
       MassBilinear[indy, indx] = MassBilinear[indy, indx] + mwi
+      MassDetBilinear[indy, indx] = MassDetBilinear[indy, indx] + dmwi  # PS TODO: sinnvoll???
```
fso42, line 1184: Mh, what would this be needed for? I.e. what is the intention of this?

avaframe/com1DFA/DFAfunctionsCython.pyx
```diff
@@ -1112,14 +1179,20 @@ def updateFieldsC(cfg, particles, dem, fields):
...
       MomBilinearX[indy, indx] = MomBilinearX[indy, indx] + mwi * ux
       MomBilinearY[indy, indx] = MomBilinearY[indy, indx] + mwi * uy
       MomBilinearZ[indy, indx] = MomBilinearZ[indy, indx] + mwi * uz
 
   for i in range(ncols):
     for j in range(nrows):
       m = MassBilinear[j, i]
+
+      #PS: TODO: sinnvoll??
+      DMDet[j, i] = DMDet[j, i] + MassDetBilinear[j, i]
```
fso42, line 1194: same here?

avaframe/com1DFA/com1DFA.py
```diff
@@ -1098,6 +1098,12 @@ def initializeSimulation(cfg, outDir, demOri, inputSimLines, logName):
     entreainableMass = np.nansum(fields["entrMassRaster"] * dem["areaRaster"])
     log.info("Mass available for entrainment: %.2f kg" % (entreainableMass))
 
+    log.debug("Initializing detrainment area")
+    detRaster, reportAreaInfo = initializeDetrainment(
+        cfgGen, dem, simTypeActual, inputSimLines["resLine"], reportAreaInfo, thresholdPointInPoly
+    )
+    fields["detRaster"] = detRaster
+
```
fso42, line 1106: Since we want the detrainment to be coupled to the resistance areas, would combining it with the resistance initialization make more sense?

avaframe/com1DFA/com1DFA.py
```diff
@@ -1582,6 +1590,51 @@ def initializeMassEnt(dem, simTypeActual, entLine, reportAreaInfo, thresholdPoin
...
+        resistanceArea = resLine["fileName"]
+        log.info("Initializing detrainment (resistance) area: %s" % (resistanceArea))
+        log.info("Detrainment (Resistance) area features: %s" % (resLine["Name"]))
+        resLine = geoTrans.prepareArea(resLine, dem, thresholdPointInPoly)
+        mask = resLine["rasterData"]
+        detRaster = K * mask
+        reportAreaInfo["detrainment"] = "Yes"
+    else:
+        detRaster = np.zeros((nrows, ncols))
+        reportAreaInfo["detrainment"] = "No"
```
fso42, line 1634:
```suggestion
    K = cfg.getfloat("detK")
    DetBool = cfg.getboolean("DetBool")
    
    # read dem header
    header = dem["originalHeader"]
    ncols = header["ncols"]
    nrows = header["nrows"]
    
    if simTypeActual in ["entres", "res"] and DetBool:
        resistanceArea = resLine["fileName"]
        log.info("Initializing detrainment (resistance) area: %s" % (resistanceArea))
        log.info("Detrainment (Resistance) area features: %s" % (resLine["Name"]))
        resLine = geoTrans.prepareArea(resLine, dem, thresholdPointInPoly)
        mask = resLine["rasterData"]
        detRaster = K * mask
        reportAreaInfo["detrainment"] = "Yes"
    else:
        detRaster = np.zeros((nrows, ncols))
        reportAreaInfo["detrainment"] = "No"
```

avaframe/com1DFA/com1DFA.py
```diff
@@ -1608,11 +1661,19 @@ def initializeResistance(cfg, dem, simTypeActual, resLine, reportAreaInfo, thres
         simulation area information dictionary completed with entrainment area info
     """
     cRes = cfg.getfloat("cRes")
+    DetBool = cfg.getboolean("DetBool")
+    DetAndRes = cfg.getboolean("DetAndRes")
```
fso42, line 1665: If we want to treat this separate and have separate functions, this logic should be treated outside this function. I.e. the algorithm should look like (for different functions):
- check if res
- then check if det on or off
- call the needed initialize functions.

OR: for only one function:

- call initialize function
- decide inside whether res and det are needed

Question is: (also @awirb ) do we want to treat it separately?

avaframe/com1DFA/com1DFA.py
```diff
@@ -2654,7 +2721,7 @@ def getSimTypeList(standardCfg, simTypeList, inputSimFiles):
     entResInfo = inputSimFiles["entResInfo"]
 
     # check if set simType is a valid option
-    validSimTypesStr = "available|null|ent|entres|res"
+    validSimTypesStr = "available|null|ent|entres|res|det"
```
fso42, line 2724: This also ties into the question, is it res with det, or is det treated separately? We need to discuss this :-)

fso42 (changes requested):

avaframe/com1DFA/DFAfunctionsCython.pyx
fso42: ok

avaframe/com1DFA/com1DFA.py
fso42: Consensus is:
no extra runtype, detrainment is an option flag of the resistance

avaframe/com1DFA/DFAfunctionsCython.pyx
```diff
@@ -436,6 +458,46 @@ cpdef (double, double) computeEntMassAndForce(double dt, double entrMassCell,
 
   return dm, areaEntrPart
 
+cpdef (double, double) computeDetMassAndForce(double dt, double detCell,
+                                              double areaPart, double uMag):
+  """ compute force component due to detrained mass
```
awirb, line 463: isn't it actually just the detrained mass that is computed? consider updating docstring

avaframe/com1DFA/com1DFA.py
PaulaSp3: So, I added the detrainment-initialization into the resistance-initialization function:
First, resistanceRaster initialized,
if detrainment is switched on: detrainment is initialized

and for only detrainment (without resistance) another flag can set resistanceRaster to zeros (only if detrainment is switched on).

avaframe/com1DFA/com1DFA.py
```diff
@@ -1621,11 +1679,27 @@ def initializeResistance(cfg, dem, simTypeActual, resLine, reportAreaInfo, thres
         # Combine constants (d, cw, sres) to one parameter cRes
         cResRaster = cRes * mask
         reportAreaInfo["resistance"] = "Yes"
+
+        if DetBool:
+            resistanceArea = resLine["fileName"]
+            log.info("Initializing detrainment (resistance) area: %s" % (resistanceArea))
+            log.info("Detrainment (Resistance) area features: %s" % (resLine["Name"]))
+            resLine = geoTrans.prepareArea(resLine, dem, thresholdPointInPoly)
+            mask = resLine["rasterData"]
```
awirb, line 1688: this is already happening outside of the if -why did add this code put it here again?

aren't just the following two lines needed?

avaframe/com1DFA/com1DFA.py
```diff
@@ -1582,6 +1586,53 @@ def initializeMassEnt(dem, simTypeActual, entLine, reportAreaInfo, thresholdPoin
 
     return entrMassRaster, entrEnthRaster, reportAreaInfo
 
+def initializeDetrainment(cfg, dem, simTypeActual, resLine, reportAreaInfo, thresholdPointInPoly):
```
awirb, line 1589: is this needed or already done in the initializeResistance? is the initialzeDetrainment called somewhere? for what flags?

avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -351,6 +351,14 @@ cpIce = 2050.0
 hRes = 10
 # combine parameters from above for resistance force to one parameter
 cRes = 0.003
+#++++++++++++ Detrainment parameter
+# Flag if snow detrainment at obstacles (in resistance area) is computed
+# detrainment can be added to resistance force (DetAndRes = True) or computed solely (DetAndRes = False)
```
awirb, line 356: update to DetWithoutRes

awirb (changes requested):

avaframe/com1DFA/DFAfunctionsCython.pyx
```diff
@@ -436,6 +458,46 @@ cpdef (double, double) computeEntMassAndForce(double dt, double entrMassCell,
...
+  cdef double dmDet = 0
+  
+  # compute detrained mass
+  if detCell > 0 and uMag > 0:
+      dmDet = - detCell / uMag * dt * areaPart
+  else:
+      dmDet = 0
+
+  if dmDet > 0 or np.isnan(dmDet):
+      dmDet = 0
```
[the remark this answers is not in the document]
PaulaSp3: it should not be met. (uMag, dt, areaPart should be > 0)

avaframe/com1DFA/com1DFA.py
PaulaSp3: not needed anymore -> I've deleted it

avaframe/com1DFA/DFAfunctionsCython.pyx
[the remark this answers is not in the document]
PaulaSp3: changed it to raise an error if the condition is met.

avaframe/com1DFA/DFAfunctionsCython.pyx
```diff
@@ -341,6 +345,21 @@ def computeForceC(cfg, particles, fields, dem, int frictType):
...
 
+      # adding detrainment analogous to entrainment
+      detCell = detRaster[indCellY, indCellX]
+      
+      if detCell > 0:
+        # compute detrained mass
+        dmDet = computeDetMassAndForce(dt, detCell, areaPart, uMag)
+        if abs(dmDet) > m:
+          # mass of a particle can't be negative
+          dmDet = - 1 * m
```
fso42, line 356:
```suggestion
          dmDet = - 1.0 * m
```

avaframe/com1DFA/com1DFA.py
```diff
@@ -1604,14 +1608,21 @@ def initializeResistance(cfg, dem, simTypeActual, resLine, reportAreaInfo, thres
...
     cResRaster : 2D numpy array
         raster of resistance coefficients
+    detRaster : 2D numpy array
+        raster of detrainment coefficients
     reportAreaInfo: dict
         simulation area information dictionary completed with entrainment area info
     """
     cRes = cfg.getfloat("cRes")
+    K = cfg.getfloat("detK")
+    DetBool = cfg.getboolean("DetBool")
```
fso42, line 1618: Please rename this to just Detrainment or similar.  There's no need to name the variable after it's type (see here: https://github.com/avaframe/AvaFrame/issues/330 ; i.e. we try to avoid this)

avaframe/com1DFA/com1DFA.py
```diff
@@ -1099,10 +1099,12 @@ def initializeSimulation(cfg, outDir, demOri, inputSimLines, logName):
     log.info("Mass available for entrainment: %.2f kg" % (entreainableMass))
 
     log.debug("Initializing resistance area")
-    cResRaster, reportAreaInfo = initializeResistance(
+    log.debug("Initializing detrainment area")
```
fso42, line 1102: I suggest to remove this log, and do this inside the Resitance function

avaframe/com1DFA/com1DFA.py
```diff
@@ -2670,6 +2700,8 @@ def getSimTypeList(standardCfg, simTypeList, inputSimFiles):
             simTypeList.append("ent")
         elif entResInfo["flagEnt"] == "No" and entResInfo["flagRes"] == "Yes":
             simTypeList.append("res")
+        elif entResInfo["flagDet"] == "Yes":
+            simTypeList.append("det")
```
fso42, line 2704: @awirb is this really needed? Because if we see detrainment as part of our resistance implementation, then we would not need this, right?

avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -351,6 +351,14 @@ cpIce = 2050.0
 hRes = 10
 # combine parameters from above for resistance force to one parameter
 cRes = 0.003
+#++++++++++++ Detrainment parameter
+# Flag if snow detrainment at obstacles (in resistance area) is computed
+# detrainment can be added to resistance force (DetAndRes = True) or computed solely (DetWithoutRes = True)
+DetBool = False
```
fso42, line 357: See above, I suggest to just use detrainment = True

avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -351,6 +351,14 @@ cpIce = 2050.0
...
+DetWithoutRes = False
```
fso42, line 358: minor, but
```suggestion
detWithoutRes = False
```

fso42 (changes requested):
Pytests fail

fso42 (changes requested):
Just to make it clear, because I think that I didn't get this across yet: please also add pytest for the new functionality

avaframe/com1DFA/com1DFA.py
PaulaSp3: I don't think that it is needed. It was from the version before, where I added a simType 'detrainment'.

fso42 (approved):

fso42:
Standardtests ok, apart from known KOT and HOF

PR #996, opened 2024-07-01
[com4FlowPy] cell_counts BUG FIX and code simplification

ahuber-bfw:
* This PR fixes the "Cell_Counts" Output Layer of com4FlowPy --> now the value in 'cell_counts' reflects the number of release pixels that reach a certain raster-cell - the extra "track-keeping" is done in a dictionary and comes at almost no additional computational cost.
* This PR also introduces a simplification to the calculation() function in flowCore.py, so that at each initialization of a flowClass.Cell() Object is only written once, instead of multiple times, depending on the "forestBool" option.

### DIFF
--- a/avaframe/com4FlowPy/flowCore.py
+++ b/avaframe/com4FlowPy/flowCore.py
@@ -336,6 +336,9 @@ def calculation(args):
     if forestBool:
         forestArray = args[11]
         forestParams = args[12]
+    else:
+        forestArray = None
+        forestParams = None
 
     z_delta_array = np.zeros_like(dem, dtype=np.float32)
     z_delta_sum = np.zeros_like(dem, dtype=np.float32)
@@ -361,6 +364,7 @@ def calculation(args):
     startcell_idx = 0
     while startcell_idx < len(row_list):
 
+        processedCells = {}  # dictionary of cells that have been processed already
         cell_list = []
         row_idx = row_list[startcell_idx]
         col_idx = col_list[startcell_idx]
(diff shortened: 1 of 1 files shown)


avaframe/com4FlowPy/flowCore.py
```diff
@@ -319,6 +319,7 @@ def calculation(args):
     # if not, wait for 30 secs and check again
     # should prevent the occurence of broken pipe errors or similar issues related
     # to RAM overflow
+    log = logging.getLogger(__name__)
```
fso42, line 322: log is never used, remove it? Or is this for a planned usage?

avaframe/com4FlowPy/flowCore.py
```diff
@@ -370,27 +375,21 @@ def calculation(args):
...
+            startcell=True,
+            FSI=forestArray[row_idx, col_idx] if isinstance(forestArray, np.ndarray) else None,
+            forestParams=forestParams,
+        )
 
+        # dictionary of all the cells that have been processed and the number of times the cell has been visited
+        processedCells[(startcell.rowindex, startcell.colindex)] = 1
+        # helper list with indices of cells, which still need to be worked
+        # this list is extended while looping through it - the calculation for one release pixel terminates
+        # when all the cells in this list have been worked / caluclated
```
fso42, line 392: -> calculated (typo)

fso42 (changes requested):

avaframe/com4FlowPy/flowCore.py
ahuber-bfw: used it for debugging and forgot to remove it

fso42 (approved):

avaframe/com4FlowPy/flowCore.py
```diff
@@ -361,6 +364,7 @@ def calculation(args):
     startcell_idx = 0
     while startcell_idx < len(row_list):
 
+        processedCells = {}  # dictionary of cells that have been processed already
```
PaulaSp3, line 367: If we used a list here, I think we could change the part where we search, if the cell already exists (line 405), that is now quite inefficient. I've tried that in the [PS_flowpy_fluxSum](https://github.com/avaframe/AvaFrame/tree/PS_flowpy_fluxSum) branch.
Just an idea for the next time :)

avaframe/com4FlowPy/flowCore.py
```diff
@@ -426,38 +423,37 @@ def calculation(args):
...
+                            startcell,
+                            FSI=forestArray[row[k], col[k]] if isinstance(forestArray, np.ndarray) else None,
+                            forestParams=forestParams,
+                                     ))
 
             z_delta_array[cell.rowindex, cell.colindex] = max(z_delta_array[cell.rowindex, cell.colindex], cell.z_delta)
             flux_array[cell.rowindex, cell.colindex] = max(flux_array[cell.rowindex, cell.colindex], cell.flux)
-            count_array[cell.rowindex, cell.colindex] += int(1)
+
+            if processedCells[(cell.rowindex, cell.colindex)] == 1:
```
PaulaSp3, line 448: why `` == 1`` ? 
``processedCells[(cell.rowindex, cell.colindex)]`` could also be larger, if the cell is hit more often (after lines 429, 430), so we don't consider these cells in the ``count_array``?

avaframe/com4FlowPy/flowCore.py
fso42: @PaulaSp3 if you have an idea like this, feel free to start a new issue from here. I added a new label called flowPyDev, just tag the issue with that...

ahuber-bfw: I just use this dictionary as a lookup table for now, i.e. every time a cell is added to the cell_list, where the actual processing is taking place, I look if this cell (col, row) is already in the dict - if not, then the cell is added with a counter of 1, if the cell has been 'visited' already this counter is incremented.

- The lookup if an element is in a list is O(n)
- The lookup if an element is in a dictionary is O(1)

so using a list for this specific purpose would be slower imo

ahuber-bfw: I also tried using a dictionary for ```cell_list```, but did not manage to produce the same results as the list implementation $\rightarrow$ But we can discuss this :)

PaulaSp3 (approved):

PR #999, opened 2024-07-02
add docu for aimec reference [ana3]

awirb:


docs/moduleAna3AIMEC.rst
```diff
@@ -7,6 +7,10 @@ For this purpose, avalanche simulation results are transformed into an avalanche
 (see :numref:`fig-aimec-comp-real`, :numref:`fig-aimec-comp-new`, :numref:`fig-aimec-domain-transfo`),
 which enables the comparison of different simulations (for example varying parameter sets,
 or performed with different models) of the same avalanche catchment, in a standardized way.
+There is also the option to compare simulation results to a reference data set, which can consist of a reference
+point, line or polygon. The simulation results are analysed regarding a chosen result variable (peak pressure, thickness or
+velocity) and a threshold of this variable and then compared towards the reference point, line or polygon.
```
fso42, line 12:
```suggestion
velocity), a threshold of this variable and then compared towards the reference point, line or polygon.
```

docs/moduleAna3AIMEC.rst
```diff
@@ -293,8 +304,27 @@ with the analysis results (see :py:func:`ana3AIMEC.ana3AIMEC.postProcessAIMEC` f
...
+This is done by identifying for each L coordinate (across thalweg) the corresponding S coordinate (along thalweg) for
+the point, line or polygon feature) (see :py:func:`ana3AIMEC.aimecTools.computeRunoutLine`).
+In case of the polygon the S coordinate furthest in thalweg direction is chosen.
+Regarding the simulation results, also a runout line is derived. This is done by identifying for each simulation, the
+last point along the thalweg where the chosen threshold of the chosen ``runoutResType`` is still exceeded (similar to the
+identification of the runout point), for each L coordinate (across thalweg).
+For each simulation the derived runout line is compared to the runout line derived from the reference line or polygon, whereas
+in the case of the reference points just a distance between this point and the simulation runout point is computed.
+The derived difference measures are saved to the aimec result DataFrame, see :ref:`moduleAna3AIMEC:List of Aimec result variables`
+and several plots are created, see :ref:`moduleAna3AIMEC:Reference data set analysis plots`.
```
fso42, line 323:
```suggestion
To include a comparison to reference data sets, the flag ``includeReference`` in your local copy of ``ana3AIMEC/ana3AIMECCfg.ini``must be set to ``True``. After reading the coordinates of the point, line or polygon from the respective shp file, a raster with the same extent as the DEM is created with values indicating which cells are affected by the features (see py:func:`ana3AIMEC.ana3AIMEC.postProcessReference`).
This raster is then transformed into the thalweg following coordinate system. From the transformed rasters, a runout line/point is computed.
This is done by identifying for each L coordinate (across thalweg) the corresponding S coordinate (along thalweg) for
the point, line or polygon feature) (see :py:func:`ana3AIMEC.aimecTools.computeRunoutLine`).
In case of the polygon the S coordinate furthest in thalweg direction is chosen. A runout line is also derived from the simulation results. This is done by identifying, for each simulation, the
last point along the thalweg where the chosen threshold of the chosen ``runoutResType`` is still exceeded (similar to the
identification of the runout point), for each L coordinate (across thalweg).
For each simulation, the derived runout line is compared to the runout line derived from the reference line or polygon, whereas
in the case of the reference points, a distance between this point and the simulation runout point is computed.
The derived difference measures are saved to the aimec result DataFrame, see :ref:`moduleAna3AIMEC:List of Aimec result variables`
and several plots are created, see :ref:`moduleAna3AIMEC:Reference data set analysis plots`.
```

fso42 (dismissed):

docs/moduleAna3AIMEC.rst
awirb: I think in this case the second and doesn't belong to the listing of inputs but just adds the next step in the procedure 
so I rephrased it: "The simulation results are analysed regarding a chosen result variable (peak pressure, thickness or
velocity) and a threshold of this variable. The derived runout point and/or line is then compared to the reference point, line or polygon."

docs/moduleAna3AIMEC.rst
awirb: done

fso42 (approved):

fso42 (approved):

PR #1004, opened 2024-07-11
[doc]: Docu for detrainment

PaulaSp3:
https://docs.avaframe.org/en/ps_com1_detrainment_doc/

### DIFF
--- a/docs/com1DFAAlgorithm.rst
+++ b/docs/com1DFAAlgorithm.rst
@@ -163,6 +163,8 @@ Other particles properties are also initialized here:
     - ``ID, parentID`` - particle IDs and parentID required if splitting, merging
 
     - ``t``- corresponding time step
+    
+    - ``dmDet`` - detrained mass of particle [kg]
 
 For more details, see :py:func:`com1DFA.com1DFA.initializeParticles`.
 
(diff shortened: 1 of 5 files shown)


fso42 (changes requested):
Please also add information to the https://docs.avaframe.org/en/latest/com1DFAAlgorithm.html#take-entrainment-into-account page similar to the entrainment info. I.e. information about how to activate it and set the parameters

PaulaSp3:
@fso42  Should I also add the detrainment computation to the algorithm graph: https://docs.avaframe.org/en/latest/com1DFAAlgorithm.html ?

fso42:
> @fso42 Should I also add the detrainment computation to the algorithm graph: https://docs.avaframe.org/en/latest/com1DFAAlgorithm.html ?
> 
> And should I also add detrainment in the [Theory](https://docs.avaframe.org/en/latest/theoryCom1DFA.html), in the mass balance equations etc.?

Yes and yes. Also with a senctence like "Snow entrainment can be added to the simulation" to make clear it is an option and not computed by default

fso42 (approved):

fso42 (dismissed):

fso42 (approved):

PR #1012, opened 2024-08-20
fix issue with entrainment from shapefile; fixes #1009

### LINKED ISSUE #1009: Handle default entrainment better
The 30cm -> CT
Currently setting nothing in shpfile leads to changed indicatior even if default from cfg file is used...

fso42:
Gives D default indicator if now entTh is given in shapefile and default is being used


awirb (dismissed):
I would only suggest to move in the com1DFACfg.ini file the entThIfMissing parameter up to entThFromShp - so that it is more obvious this will be used compared to relTh where then no relTh is set set if missing in shp file

awirb (approved):

fso42:
Standardtests ok, apart from known kot and hof

Issue #872, opened 2023-06-22
Report problem Scenario vs Featurename


fso42:
Report has the last shapefilename instead of the matching one for the scenario name

reported by Christian T.

fso42:
QGis problem?

fso42:
Seems to be a QGis GUI (!!) problem. 
- execute AvaFrame via QGis Gui results in report error (ie. wrong scenario name vs Release area and simname)
- execute the SAME avalanche in the corresponding QGis Shell via python -m avaframe.runCom1DFA AVANAME does NOT produce the error!!!!!!

PR #1024, opened 2024-09-16
Handle bug about simName / scenario mismatch in report (Windows QGis GUI); issue #872 [com1]

fso42:
Fixes the problem with report sinNames on Windows (#872)

This is a dirty fix, because it does not address the root cause, simply because I cannot find it. It only treats the effect...

### DIFF
--- a/avaframe/log2Report/generateReport.py
+++ b/avaframe/log2Report/generateReport.py
@@ -233,3 +233,39 @@ def writeReport(outDir, reportDictList, reportOneFile, plotDict='', standaloneRe
 
                 # Write report file
                 writeReportFile(reportD, pfile)
+
+
+def checkAndCleanReportDictOnWinIssue872(reportDictList):
+    """ Check report dict for inconsistencies in simName. Sometimes the simName is not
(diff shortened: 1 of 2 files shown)


avaframe/log2Report/generateReport.py
```diff
@@ -233,3 +233,39 @@ def writeReport(outDir, reportDictList, reportOneFile, plotDict='', standaloneRe
...
+            nameParts = simName.split('_AF_')
+        else:
+            nameParts = simName.split('_')
+
+        # This is the proper simName
+        simNameClean = nameParts[0]
+
+        if listItem['Simulation Parameters']['Release Area Scenario'] != simNameClean:
+            reportDictList[k]['Simulation Parameters']['Release Area Scenario'] = simNameClean
+            log.warning("Ran into windows QGis simName report problem, set to proper simName now")
```
awirb, line 265: suggestion, set release area scenario according to simName now? 

or any information in the warning that this concerns the release area scenarion info?

awirb (approved):

PR #1030, opened 2024-09-25
[com4FlowPy]: Add docstrings to functions

PaulaSp3:
Added docstrings to function.

- I almost didn't change existing docstrings/ documentation
- I'm not sure how to document the object methods properly (``flowClass.py``)

### DIFF
--- a/avaframe/com4FlowPy/__init__.py
+++ b/avaframe/com4FlowPy/__init__.py
@@ -1,0 +1,4 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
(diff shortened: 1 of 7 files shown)


fso42:
@PaulaSp3: This is where the docstring ends up: https://docs.avaframe.org/en/latest/_autosummary/com4FlowPy.flowClass.Cell.html#com4FlowPy.flowClass.Cell . 
Looking at it, I think if you specifiy the input after the "constructor for the cell..." bit, you should be fine....

fso42 (approved):

PR #1031, opened 2024-09-25
Update Advanced Usage chapter in docs [doc]

leon-wagner:
Added installation description for Windows, reorganized some (so e.g. Configuration is under Advanced Usage in docs, as well as moving update to the bottom), adjusted :refs: so they properly match

https://docs.avaframe.org/en/wininstalldoc/

### DIFF
--- a/docs/developinstallwin.rst
+++ b/docs/developinstallwin.rst
@@ -1,0 +1,129 @@
+Advanced Installation (Windows)
+------------------------------
+
+This is a quick guide on how to install AvaFrame
+and the required dependencies on your machine. AvaFrame is developed on **Linux
+machines** (Ubuntu/Manjaro/Arch) with recent Python versions > 3.8.
+These instructions assume you are familiar with working in a terminal. This
+guide is described for **Windows**. For *Linux*, see :ref:`developinstall:Advanced Installation (Linux)`.
+
+Requirements
+^^^^^^^^^^^^
+
+Install `git <https://github.com/git-guides/install-git>`_ and python, we
+suggest to work with miniconda/anaconda. For installation see `miniconda
+<https://docs.conda.io/en/latest/miniconda.html>`_ or
+`anaconda <https://docs.anaconda.com/anaconda/install/linux/>`_.
+
+Install `Microsoft C++ compiler <https://wiki.python.org/moin/WindowsCompilers>`_.
+Follow the installation steps for the version corresponding to the installed python version.
+
+Setup AvaFrame
+^^^^^^^^^^^^^^
+
+Open conda shell and create a new `conda environment
+<https://conda.io/projects/conda/en/latest/user-guide/concepts/environments.html>`_
+for AvaFrame, activate it and install pip, numpy and cython in this environment::
+
+  conda create --name avaframe_env
+  conda activate avaframe_env
+  conda install pip numpy cython
+
+Clone the AvaFrame repository (in a directory of your choice: [YOURDIR]) and change into it::
+
+  cd [YOURDIR]
+  git clone https://github.com/avaframe/AvaFrame.git
+  cd AvaFrame
+
+Compile the cython com1DFA part::
+
+   python setup.py build_ext --inplace
+
+.. Warning::
+   You will have to do this compilation every time something changes in the cython code. We also suggest
+   to do this everytime updates from the repositories are pulled.
+
+   **Before** compilation in Windows, make sure to delete ``AvaFrame/build`` directory, in addition to any .pyd, .c, and
+   .pycache files in ``AvaFrame/avaframe/com1DFA``
+
+Install avaframe and its requirements by **either** doing::
+
+  pip install -e .
+
+or if this fails (see `github issue 986 <https://github.com/avaframe/AvaFrame/issues/986>`_), do::
+
+  python setup.py develop
+
+This installs avaframe in editable mode, so every time you import avaframe the
+current (local) version will be used.
+
+Test it by starting ``python`` and do an ``import avaframe``. If no error comes
+up, you are good to go.
+
+If wanted: Integrate conda environment into editor of your choice (i.e. PyCharm, Spyder or any other)
+
+Head over to :ref:`gettingstarted:First run` for the next steps.
+
+Using QGIS from Conda
+^^^^^^^^^^^^^^
+It is possible to have the script installation and the plugin installed on your machine at the same time.
+However, be aware: Depending on your mode of QGis installation (direct installer, conda, OSGeo4W...) your script AvaFrame installation might be overruled.  
+
+The steps below install QGis and the Connector plugin in a separate *conda* environment. If the steps above have been followed, this method will result in two script installations where one will be used by QGis and the other can be worked on separately, preventing any overlap.
+
+If you have another QGis installation somewhere, make sure that the AvaFrame plugin is not installed.
+To create and activate a new environment, which will contain your installation of QGis open your conda terminal and run::
+
+  conda create -n qgis_latest
+  conda activate qgis_latest
+
+To **install the latest QGIS version**, use::
+
+  conda install qgis --channel conda-forge
+
+.. Note::
+   Conda will always try to install the latest version. If you want to use another version, you need to specify it, e.g.:
+
+  ``conda install qgis=3.34.11 --channel conda-forge``
+
+Now install the avaframe requirements::
+
+  conda install pip numpy cython
+
+Now find the directory where your environment is located. The path should be something like:
+``C:\Users\USER\miniconda3\envs\qgis_latest``. Change into it::
+
+  cd \miniconda3\envs\qgis_latest
+
+Clone the AvaFrame repository and change into it::
+
+  git clone https://github.com/avaframe/AvaFrame.git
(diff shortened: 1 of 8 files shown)


leon-wagner:
Find solution to QGIS and script overwrite issue and add to docs?

docs/developinstallwin.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Compile the cython com1DFA part::
+
+   python setup.py build_ext --inplace
+
+.. Warning::
+   You will have to do this compilation every time something changes in the cython code. We also suggest
+   to do this everytime updates from the repositories are pulled.
+
+   **Note**: After (re)compilation in Windows, delete AvaFrame/build directory, in addition to any .pyd, .c, and .pycache files in AvaFrame/avaframe/com1DFA
```
fso42, line 46:
```suggestion
   **Note**: For compilation in Windows, make sure to delete AvaFrame/build directory, in addition to any .pyd, .c, and .pycache files in AvaFrame/avaframe/com1DFA
```

docs/developinstallwin.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Test it by starting ``python`` and do an ``import avaframe``. If no error comes
+up, you are good to go.
+
+If wanted: Integrate conda environment into Editor of your choice (i.e. PyCharm, Spyder or any other)
+
+Head over to :ref:`gettingstarted:First run` for the next steps.
+
+NOTE:
+  Installing AvaFrame on Windows via QGIS plugin writes AvaFrame to system-wide python. This will override any avaframe installation in conda environment!
```
fso42, line 67: Mh, this is not entirely true but depends on the mode of installation. There is no such thing as system-wide python on Windows, only python installations that might be accessed/shared. I suggest rewriting it to something like:
```suggestion
  Installing AvaFrame on Windows via QGIS plugin might overrule your script AvaFrame installation! This depends on your mode of installation (conda, OSGeo4W, direct installer...).
```

docs/moduleCom1DFA.rst
```diff
@@ -52,6 +52,7 @@ projection:
 
 * release area scenario as (multi-) polygon shapefile (in Inputs/REL; multiple features are possible)
 
+  - the release area polygon should not contain any "holes" or inner rings
```
fso42, line 55:
```suggestion
  - the release area polygon must not contain any "holes" or inner rings
```
This is good, but please move it to a separate point and write : Any shapefile with polygons must not....

fso42 (changes requested):

docs/developinstallwin.rst
```diff
@@ -0,0 +1,147 @@
...
+
+If wanted: Integrate conda environment into editor of your choice (i.e. PyCharm, Spyder or any other)
+
+Head over to :ref:`gettingstarted:First run` for the next steps.
+
+Using QGIS from Conda
+^^^^^^^^^^^^^^
+It might be wanted to both have a script installation and the plugin installed on your machine at the same time.
+Unfortunately, installing AvaFrame on Windows via QGIS plugin might overrule your script AvaFrame installation, causing
+issues when trying to work on both versions in parallel. This depends on your mode of installation (direct installer, conda, OSGeo4W...).
```
fso42, line 71: More direct and concise:
```suggestion
It is possible to have the script installation and the plugin installed on your machine at the same time.
However, be aware: Depending on your mode of QGis installation (direct installer, conda, OSGeo4W...) your script AvaFrame installation might be overruled.  
```

docs/developinstallwin.rst
```diff
@@ -0,0 +1,147 @@
...
+
+Below, we have compiled a **short guide**, describing a way to install QGIS and the Connector plugin in a separate
+*conda* environment. This method will result in two script installations (if the steps above have been followed),
+where one will be used by QGIS and the other may be worked on separately, preventing any overlap.
```
fso42, line 75:
```suggestion
The steps below install QGis and the Connector plugin in a separate *conda* environment. If the steps above have been followed, this method will result in two script installations where one will be used by QGis and the other can be worked on separately, preventing any overlap.
```

docs/developinstallwin.rst
```diff
@@ -0,0 +1,147 @@
...
+
+Firstly, make sure that if you have another QGIS installation somewhere, that the AvaFrame plugin is not installed.
+Next, create and activate a new environment, which will contain your QGIS installation. The name of the environment
+can be anything, but since we will install the latest version of QGIS, we will call it *qgis_latest* in this example.
+Open your conda terminal and use::
```
fso42, line 80:
```suggestion
If you have another QGis installation somewhere, make sure that the AvaFrame plugin is not installed.
To create and activate a new environment, which will contain your installation of QGis open your conda terminal and run::
```

docs/developinstallwin.rst
```diff
@@ -0,0 +1,147 @@
...
+
+  conda create -n qgis_latest
+  conda activate qgis_latest
+
+To **install the latest QGIS version**, use::
+
+  conda install qgis --channel conda-forge
+
+.. Note::
+  Because conda will always try to install the latest version, if you want to use another version, you need to specify it, e.g.:
```
fso42, line 90:
```suggestion
   Conda will always try to install the latest version. If you want to use another version, you need to specify it, e.g.:
```

docs/developinstallwin.rst
```diff
@@ -0,0 +1,147 @@
...
+  pip install avaframe
+
+Now find the directory where your environment is located. The path should be something like:
+``C:\Users\USER\miniconda3\envs\qgis_latest``. Change into it::
+
+  cd \miniconda3\envs\qgis_latest
+
+.. Note::
+  **Tip:** To find the path, you can do ``pip show avaframe`` and copy the
+  first part which shows your environment directory
```
fso42, line 106: I think we could leave this out and just say: use a directory that's different then your script installation from above, right? Since then we would not touch any conda env directories. The requirements should come through the commands below anyways...

docs/developinstallwin.rst
```diff
@@ -0,0 +1,147 @@
...
+Compile the cython com1DFA part and create an editable avaframe version in your environment::
+
+  python setup.py build_ext --inplace
+  pip install -e .
+
+To **run QGIS**, in the terminal, activate the environment (if not activated already) and run the ``qgis`` command::
+
+  conda activate qgis_latest
+  qgis
+
```
fso42, line 128: Mh, I would only write 'Run qgis: ' 

If people followed the guide, the environment should be activated. If not, they installed AvaFrame into a systemwide directory, leading to problems...

docs/developinstallwin.rst
```diff
@@ -0,0 +1,147 @@
...
+  conda update qgis -c conda-forge
+
+**Final comments:**
+
+Be aware that the QGIS Conda package does not provide the same user experience as the QGIS installer version.
+For example, there are no desktop icons or file association, and it does not include GRASS and SAGA, etc..
+
+On the other hand, the QGIS installation through Conda will share user configurations, installed plugins,
+with any other QGIS installation on your system.
```
fso42, line 147: I would leave this out, since I'm not sure that is true for all setups...

fso42 (changes requested):

docs/developinstallwin.rst
leon-wagner: I am not sure about this one.. I believe it's important to install the requirements before in the environment right? At least we need pip numpy cython. Or will the cloned AF be able to use dependencies outside the environment?

docs/developinstallwin.rst
leon-wagner: Sure, for the installation it's not needed to activate the env again. I was thinking for people just to refer back on how to open qgis again, after pc restart etc., to make sure again to activate the env. Could alternatively move the info downwards..

docs/developinstallwin.rst
leon-wagner: This info is from [here](https://plugins.qgis.org/planet/tag/conda/). True, could leave this out though, as I am not completely sure either how QGIS works there..

docs/developinstallwin.rst
fso42: That is a good point, installing avaframe however is not a good idea. Please switch `pip install avaframe` to `conda install pip numpy cython`. The rest of the requirements will be met via the pip install lower down.

docs/developinstallwin.rst
fso42: In my experience it is better to keep the explanations as simple as possible: having the activate a second time here could lead to a 'double stacked' environment, which should not be a problem, but could be.... In the context/flow of this explanation please leave the activate out here. 
And furthermore: I'm hoping that anyone who is attempting this has at least the presence to get familiar with conda in case they don't know it already. So they should be aware of the fact that an environment has to be activated before they can use it again. Especially since QGis won't be found without the env...

docs/developinstallwin.rst
fso42: The info you linked is from 2019 and as you/we are not completely sure how this is handled on current version, please just leave it out...

fso42 (changes requested):

fso42 (approved):

PR #1041, opened 2024-11-04
revers long long vs long again; adjust setup.py and releaseMany [com1]

fso42:


fso42:
Standardtests ok, apart from know avaKot

PR #1042, opened 2024-11-11
com6RockAvalanche: add scarp.py, runCom6Scarp.py & scarpCfg.ini [com6]

dwolfsch:
- [x] move Readme to doc -> @fso42 
- [x] Use avadir setup for Input (as raster/shape) similar to original input mu/xi etc -> @dwolfsch 
- [x] Get feature information from shapefile
https://docs.avaframe.org/en/scarp/moduleCom6RockAvalanche.html


avaframe/com6RockAvalanche/scarpCfg.ini
```diff
@@ -0,0 +1,57 @@
...
+# This is used only if 'shapefile' is not provided.
+# 
+# For 'plane' method: Define multiple planes by listing xseed, yseed, zseed, dip, slope for each plane.
+# Example for three planes:
+# features = x1,y1,z1,dip1,slope1,x2,y2,z2,dip2,slope2,x3,y3,z3,dip3,slope3
+#
+# For 'ellipsoid' method: Define multiple ellipsoids by listing x_center, y_center, max_depth, semi_major_axis, semi_minor_axis for each ellipsoid.
+# Example for two ellipsoids:
+# features = x1,y1,depth1,a1,b1,x2,y2,depth2,a2,b2
+features = 68247,299770,920,0,0,68111,299832,850,90,50,68340,299799,850,270,50
```
fso42, line 43: Would this fit better in the INPUT section?

avaframe/runCom6Scarp.py
```diff
@@ -0,0 +1,10 @@
+#RunScarp.py
```
fso42, line 1: Please look at one of our runscripts, for example AvaFrame/avaframe/runIn1RelInfo.py and copy that structure (with the main etc). This is necessary to be easily callable from the QGis connector. 

I also suggest using a different way of getting the config file. i.e. instead of giving it directly, import the module scarp and then load the config. This way, the normal logic of possible usage of local_cfg.ini etc is observered. If you are not sure about what is needed to do this, let me know and I'm happy to help.

fso42 (commented):
Hi @dwolfsch 
Thank you for your PR. 
I added the first round of review below, but it already looks good. 
A general point:  
A README or a section in the documentation would be perfect. This should explain the basic usage to make it easier for users what happens and what is needed to run this tool. 
If you would rather not touch the documentation, use a README at first, we can then transfer this to the docs.

avaframe/com6RockAvalanche/scarp.py
```diff
@@ -0,0 +1,312 @@
...
+    elif method == 'ellipsoid':
+        scarpData = calculateScarpWithEllipsoids(elevData, periData, elevTransform, features)
+    else:
+        raise ValueError("Unsupported method. Choose 'plane' or 'ellipsoid'.")
+
+    hRelData = elevData - scarpData
+    saveRaster(elevScarp, scarpData, elevTransform, elevCRS)
+    saveRaster(hRelease, hRelData, elevTransform, elevCRS)
+
+def readPerimeter(perimeterInput, perimeterShapefilePath, elevTransform, elevShape, elevCRS):
```
fso42, line 113: elevCRS is unused, is this on purpose?

fso42 (approved):

fso42:
Standardtests ok

PR #1045, opened 2024-11-19
Fix no points

awirb:
if features have no points but are still available in shp file - remove these features log warning that these area ignored


avaframe/in2Trans/shpConversion.py
```diff
@@ -93,72 +93,79 @@ def SHP2Array(infile, defname=None):
...
-                    rho = value
-                if (name == 'sks'):
-                    sks = value
-                if (name == 'iso'):
-                    iso = value
-                if (name == 'layer'):
-                    layerN = value
-            # if name is still empty go through file again and take Layer instead
-            if ((type(layername) is bytes) or (layername is None)):
+        if len(pts) > 0:
```
fso42, line 100: Maybe I'm thinking too simple, but could we just to a 
if len <= 0:
  continue? 
And don't do all the other bookkeeping? 
I.e. if an item without points in encountered, we just don't do anything? I'm pretty sure the idlist could be built inside the first loop. 


If this is not possible for some reason, then I suggest trying to split the nesting...

fso42 (changes requested):
The test shapefile is missing

avaframe/in2Trans/shpConversion.py
awirb: use continue and loop also over sf.records() in first loop to access records oid for idList

fso42 (approved):

PR #1060, opened 2025-01-07
fix com2AB multiple lines issue [com2]

### LINKED ISSUE #1057: com2AB Problem with multiple lines but no name attribute
Problem:
- Shapefile with multiple lines
- But no attribute 'Name'
- Only one result comes back since we only have one "default" name.

Expected:
- All results should come back with a generic name i.e. line1 line2 etc

leon-wagner:
fixes #1057


fso42 (changes requested):
Please adjust the pytest to reflect the "line" name

fso42 (approved):

fso42:
Standardtests ok, apart from know Kot and Hof

Issue #1061, opened 2025-01-09
[com4FlowPy]: fix zDeltaSum calculation to be in line with cellCount calculation


ahuber-bfw:
* The cellCounts (countArray) now *correctly* contain the number of release cells, that route flux through a given cell
* However, the way it is implemented now we are not able to calculate average zDelta by simply dividing the zDeltaSum with the cellCounts result layers (because they are not summed up in the same way here).

ahuber-bfw:
> Do you mean that it would make sense to take the maximum value of zDelta in every cell for every path (per startcell) and then the sum of these zDeltaMax values over all paths? I'm not so sure, what we want as result for zDeltaSum...

yep, that would IMO make the most sense, because then we can interpret zDeltaSum/cellCounts as some sort of "average energy/velocity" at a given point. The way it is implemented now zDeltaSum can hardly be interpreted at all ...

PR #1067, opened 2025-01-17
[com4FlowPy]: new computation of zDeltaSum

PaulaSp3:
zDeltaSum is now computed in that way: 
- for every path (startcell): the maximum of zDelta in every cell
- for the output raster: the sum of all paths

-> now especially, the average of zDelta as zDeltaSum / cellCounts is (easier) to interpret

see issue #1061


ahuber-bfw (approved):
looks good from my side

PR #1084, opened 2025-02-07
[com4] Further mass redistribution/correction modifications

theweathermanda:
Further changes to the mass redistribution/correction procedure in com4FlowPy/flowClass.py:
- Update criterion for the mass redistribution/correction step to be 'self.dist >= threshold' instead of 'self.dist > threshold'. Else, cells with values equal to the specified threshold aren't dealt with
- Add a second adjustment step for scenarios with a net gain of mass due to rounding errors in above arithemetic operations
- Update row_local, col_local to be cells with self.dist >= threshold, consistent with above

I recently uncovered the mass conservation issues in the FlowPy version used in AutoATES prior to finding this updated repo, and in tests with several print statements tracking the values of self.dist and self.flux through this procedure, I found that these changes should lead to mass conservation, notwithstanding any 'mass loss' due to deposition, which you have included here.

### DIFF
--- a/avaframe/com4FlowPy/flowClass.py
+++ b/avaframe/com4FlowPy/flowClass.py
@@ -391,11 +391,10 @@ class Cell:
             # local flux redistribution to eligible child cells
             self.dist[self.dist >= threshold] += mass_to_distribute / count
             self.dist[self.dist < threshold] = 0
-        if np.sum(self.dist) < self.flux and count > 0:
-            # correction/flux conservation for potential rounding losses
-            self.dist[self.dist >= threshold] += (self.flux - np.sum(self.dist)) / count
-        if np.sum(self.dist) > self.flux and count > 0:
-            # correction/flux conservation for potential rounding gains
+        if np.sum(self.dist) != self.flux and count > 0:
+            # correction/flux conservation for potential rounding losses or gains
+            # (self.flux - np.sum(self.dist)) will either be negative or positive 
+            # depending on the direction of the rounding error
             self.dist[self.dist >= threshold] += (self.flux - np.sum(self.dist)) / count
         if count == 0:
             # if all child cells are below flux_threshold, the flux is deposited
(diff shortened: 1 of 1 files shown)


fso42:
@theweathermanda : Thank you for your PR. Either @PaulaSp3 and/or @ahuber-bfw will review it.

avaframe/com4FlowPy/flowClass.py
```diff
@@ -389,16 +389,19 @@ def calc_distribution(self):
...
-            self.dist[self.dist > threshold] += mass_to_distribute / count
+            self.dist[self.dist >= threshold] += mass_to_distribute / count
             self.dist[self.dist < threshold] = 0
         if np.sum(self.dist) < self.flux and count > 0:
             # correction/flux conservation for potential rounding losses
-            self.dist[self.dist > threshold] += (self.flux - np.sum(self.dist)) / count
+            self.dist[self.dist >= threshold] += (self.flux - np.sum(self.dist)) / count
+        if np.sum(self.dist) > self.flux and count > 0:
+            # correction/flux conservation for potential rounding gains
+            self.dist[self.dist >= threshold] += (self.flux - np.sum(self.dist)) / count
```
ahuber-bfw, line 399: thx, correction for rounding gains also was missing, maybe you can simplify by handling both cases (rounding gains, rounding losses) in a single statement
```suggestion
        if np.sum(self.dist) != self.flux and count > 0:
            # correction/flux conservation for potential rounding losses or gains
            # (self.flux - np.sum(self.dist)) will either be negative or positive 
            # depending on the direction of the rounding error
            self.dist[self.dist >= threshold] += (self.flux - np.sum(self.dist)) / count
```

ahuber-bfw (approved):
Thx for pointing this out and the patch.

avaframe/com4FlowPy/flowClass.py
fso42: Hi @theweathermanda : is the suggestion ok for you? Then you could commit it and I'll merge this PR...

theweathermanda: @fso42 Added a commit below for this change. Thanks!

fso42:
@theweathermanda Thanks again. I will include this in the 1.12 release (the next 1.11 release will only include updates on benchmarks).

PR #1088, opened 2025-02-13
[sys] Update ALL benchmarks

fso42:
This updates all Benchmark with the code changes from version 1.10


awirb (approved):
update docu com1DFA - change reference to com1DFAOrig to an archived version of the docu?

PR #1095, opened 2025-02-21
[com4] use raster functions (for .asc and .tif) from AvaFrame

PaulaSp3:
- use the raster - functions from ``avaframe.in2Trans.rasterUtils`` and ``avaframe.in1Data.getInput`` instead of ``avaframe.com4FlowPy.raserIo``
- ``avaframe.com4FlowPy.raserIo``is not needed anymore (can be deleted, not done yet..)
- the only differences I recognized:
  - when writing rasterData now the **noData values** (before: nodata: -9999) -> but I still could change the if required
  - we flip upside down the rasters when reading in and writing, when doing that consistently, I could not recognize a problem yet.


avaframe/com4FlowPy/splitAndMerge.py
```diff
@@ -42,7 +42,8 @@ def tileRaster(fNameIn, fNameOut, dirName, xDim, yDim, U, isInit=False):
     #    os.makedirs(dirName)
 
     # largeRaster, largeHeader = iof.f_readASC(fNameIn, dType='float')
-    largeRaster, largeHeader = io.read_raster(fNameIn)
+    largeData = IOf.readRaster(fNameIn)
```
ahuber-bfw, line 45: I think we need to include `noDataToNan=False` here, otherwise we get problems with some existing code that is not able to handle `np.nan`

```suggestion
    largeData = IOf.readRaster(fNameIn, noDataToNan=False)
```

ahuber-bfw: see commit below

avaframe/com4FlowPy/com4FlowPy.py
```diff
@@ -604,34 +598,25 @@ def checkConvertReleaseShp2Tif(modelPaths):
...
-            _errMsg = "using release area in '.shp' format currently only supported in combination with '.asc' DEMs"
-            log.error(_errMsg)
-            raise ValueError(_errMsg)
-        else:
-            _errMsg = f"file format '{ext}' for DEM is not supported, please provide '.tif' or '.asc'"
-            log.error(_errMsg)
-            raise ValueError(_errMsg)
-
-        dem['originalHeader'] = demHeader
+        dem = IOf.readRaster(modelPaths["demPath"])
```
PaulaSp3, line 603: @ahuber-bfw , here we co not need ``noDataToNan=False``, right?

avaframe/com4FlowPy/com4FlowPy.py
```diff
@@ -604,34 +598,25 @@ def checkConvertReleaseShp2Tif(modelPaths):
...
+        demHeader = dem["header"]
+        dem['originalHeader'] = demHeader # TODO: why?
```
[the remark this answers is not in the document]
PaulaSp3: As I understand right, this is necessary for gT.polygon2Raster ?

ahuber-bfw (approved):
Looks good :) - I made some comments and pushed some minor changes --> overall this looks good and works for all the cases I tested it with:
* `useCustomPaths = False`
* `useCustomPaths = True` with release area as `.tif` und `.shp`
* with "clipped" DEM (i.e. there are nodata_values in some areas of the dem extent)
* with and without `forest` and `infra`

only thing I didn't test was the variable Input Layers (`variableAlpha`, `variableUmaxLim`, `variableExponent`) - but this should also be fine based on the other tests.

thx for the effort @PaulaSp3

PR #1103, opened 2025-03-07
[in3] Problem with non-closed shape polygons 

awirb:
add enpoint=startpoint  to idealized release areas and in getRelease function when writing to shp file


PaulaSp3 (approved):

fso42:
Standardtest ok (by AW)

PR #1105, opened 2025-03-12
[com1] add option for resistance

awirb:
This adds:
- detrainment in resistance areas if v and th are below thresholds
- disables resistance if v and th are above thresholds (thresholds are really high atm, should basically never apply; added for testting)


fso42 (changes requested):
- [ ] Change snow slide res parameter
- [ ] Apply qtly suggestions about is not None

avaframe/com1DFA/com1DFA.py
```diff
@@ -1628,19 +1631,27 @@ def initializeResistance(cfg, dem, simTypeActual, resLine, reportAreaInfo, thres
...
     if simTypeActual in ["entres", "res"]:
+        ResModel = cfg["ResistanceModel"].lower()
+        if ResModel == "default":
+            cRes = cfg.getfloat("cResH")
+            if detrainment:
+                log.info('Resistance model default: if FT < %.2f OR FV < %.2f detrainment (detK %.2f) else increased friction (cResH %.2f) and no effect if FT > %.2f OR FV > %.2f' %
+                         (cfg.getfloat('forestThMin'), cfg.getfloat('forestVMin'), K, cfg.getfloat('cResH'),
+                          cfg.getfloat('forestThMax'), cfg.getfloat('forestVMax')))
+            else:
+                log.warning('Resistance model default chosen, but detrainment is deactivated apply increased friction (cResH %.2f) everywhere within resistance area' % cfg.getfloat('cResH'))
```
fso42, line 1649: I don't think this is necessary, I suggest removing it

avaframe/com1DFA/com1DFA.py
```diff
@@ -1762,6 +1768,7 @@ def DFAIterate(cfg, particles, fields, dem, inputSimLines, outDir, cuSimName, si
     massEntrained = []
     massDetrained = []
     massTotal = []
+    pfvTimeMax = []
```
fso42, line 1771: What is the purpose of this variable? And the difference to pfv?

avaframe/com1DFA/com1DFA.py
```diff
@@ -2213,6 +2228,17 @@ def computeEulerTimeStep(cfg, particles, fields, zPartArray0, dem, tCPU, frictTy
...
     """
+
+    # update cRes and detK rasters according to thresholds of FV and FT
+    particles['tPlot'] = particles['tPlot'] + 1
+    # only if entres or res sim and detrainment is used
+    if cfg['simTypeActual'] in ['entres', 'res'] and (cfg['ResistanceModel'].lower() == 'default' and cfg.getboolean('detrainment')):
+        # update resistance area fields using thresholds
+        fields = com1DFATools.updateResCoeffFields(fields, cfg, float(particles['t']), dem)
+        # plot
+        #outCom1DFA.plotResFields(fields, cfg, particles['tPlot'], dem, particles['mTot'])
```
fso42, line 2239: please remove commented out or put behind debugPlot flag

avaframe/com1DFA/com1DFATools.py
```diff
@@ -380,3 +382,63 @@ def checkCfgInfoType(cfgInfo):
         raise AssertionError(message)
 
     return typeCfgInfo
+
+
+def updateResCoeffFields(fields, cfg, t, dem):
```
fso42, line 387: t and dem are not in the docstring

avaframe/com1DFA/com1DFATools.py
```diff
@@ -380,3 +382,63 @@ def checkCfgInfoType(cfgInfo):
...
+    """ update fields of cRes and detK, coefficients of resistance parameter and detrainment parameter
+        according to the thresholds of FV and FT
```
fso42, line 389: set indent as the other docstrings

avaframe/com1DFA/com1DFATools.py
```diff
@@ -380,3 +382,63 @@ def checkCfgInfoType(cfgInfo):
...
+        fields['cResRasterOrig'] = cResOrig
+        fields['detRasterOrig'] = detOrig
+        log.info('Resistance area removed %d cells because FV or FT exceeded %.2f ms-1, %.2f m' % (lTh, vMax, thMax))
+        outFileRes = pathlib.Path(cfg['avalancheDir'], 'Outputs', 'com1DFA', 'reports',
+                                  ('resArea_t%.2f' % t))
+        IOf.writeResultToRaster(dem["originalHeader"], fields['cResRasterOrig'], outFileRes, flip=True)
+
+    # update fields dictionary
+    fields['cResRaster'] = cResRaster
+    fields['detRaster'] = detRaster
```
fso42, line 441: These end up in the output, right? Then please add them to https://docs.avaframe.org/en/latest/moduleCom1DFA.html#output

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -149,6 +149,7 @@ def checkResType(fullCfg, section, key, value):
             "TA",
             "particles",
             "dmDet",
+            "FTDet"
```
fso42, line 152: Add to output info https://docs.avaframe.org/en/latest/moduleCom1DFA.html#output and to com1Cfg.ini info (on resType parameter)

fso42 (changes requested):

PaulaSp3 (approved):
I've only tested the changes with doing a few test runs.
@fso42 and @awirb, if I also should check the code, please tell me.

avaframe/com1DFA/com1DFA.py
awirb: I find this a quite helpful information in the log

avaframe/com1DFA/com1DFA.py
awirb: it is an array of the max value for each time step - needed for the new mass plot better naming would be fvTime Max so it is a scalar max value over all cells but for each time step

avaframe/com1DFA/com1DFATools.py
awirb: not to be exported it is just to store them during the computation so that they can be used during the computation

avaframe/com1DFA/deriveParameterSet.py
awirb: added int the outputs

avaframe/com1DFA/com1DFA.py
fso42: If you want to keep it, then please drop it to debug. In my opinion this is unnecessary, since it is described in the documentation as well as written in the ini file and additionally these values are also written in the log earlier on. And if someone changes the detrainment  in the ini file, they are aware of it...
If we start writing the configuration setup as separate log messages as well, then it's going to be a very messy log....

fso42 (approved):

fso42:
Standardtests ident apart from the res ones:
big changes on avaInclinedPlane and avaParabola (especially the mass)
minimal change on avaHof

Benchmark results will be updated with 1.13.1

PR #1123, opened 2025-04-30
[com1] Add comment, fix small bug

RolandFischbacher:


avaframe/in3Utils/geoTrans.py
```diff
@@ -171,7 +171,7 @@ def projectOnGrid(x, y, Z, csz=1, xllc=0, yllc=0, interp="bilinear", getXYField=
 
 def resizeData(raster, rasterRef):
     """
-    Reproject raster on a grid of shape rasterRef
+    Reproject raster on a grid of shape rasterRef, raster adapts cellsize and extend of rasterRef
```
PaulaSp3, line 174:
```suggestion
    Reproject raster on a grid of shape rasterRef, raster adapts cellsize and extend of rasterRef!
```

PaulaSp3 (approved):

avaframe/com1DFA/deriveParameterSet.py
```diff
@@ -885,12 +885,13 @@ def checkExtentAndCellSize(cfg, inputFile, dem, fileType):
...
 
     rT = float(cfg["GENERAL"]["resizeThreshold"])
+    cT = float(cfg["GENERAL"]["meshCellSizeThreshold"])
 
     diffX0, diffX1, diffY0, diffY1 = checkSizeExtent(inputField, demHeader, inputFile, fileType, rT)
 
     # check if identical extent, if so use unchanged
     if (
-        np.allclose([diffX0, diffY0], [diffY1, diffX1])
+        np.allclose([diffX0, diffY0, diffX1, diffY1], [0, 0, 0, 0], atol=cT)
```
fso42, line 894: I am not sure this is what we want? 
So we want to see whether the extent is the same, why use the meshCellSizeThreshhold then?

fso42 (changes requested):

avaframe/com1DFA/deriveParameterSet.py
RolandFischbacher: The problem with the previous version is that the condition is true when the number of rows, columns, and cell size are the same (although the raster could be shifted in comparison to the reference raster). So, if only xllcenter and yllcenter are different, no remeshing will be done, and thus the test fails.

I thought that remeshing should be done if the differences in the lower left and upper right corner coordinates are not zero, using meshCellSizeThreshold as the threshold below which no remeshing should be done.

fso42:
Standardtests ident

PR #1149, opened 2025-06-25
Add initial com8MotPSA setup

fso42:
Might not run yet


fso42:
Standardtests ok, apart from know res issues

PR #1159, opened 2025-07-08
[com4]: Add result file: minimum of travel length and travel angle

PaulaSp3:
-


ahuber-bfw (changes requested):
Thx for the PR,

* the new outputs (`travelLengthMin`, `fpTravelAngleMin`) are producing expected results!
* In release areas the use of `min` when overlaying results from different paths leads to the effect, that (probably all) release cells will have values of `0` for both the `travelAngle` and the `travelLength` output --> this has to be interpreted accordingly in the results

two minor change requests:
* maybe we can also initialize the `travelLenghtMax` and `fpTravelAngleMax` with `np.ones_like() * -9999` - so behaviour is the same for min and max outputs
* if it's easy to implement, we could only execute code parts if the respective output layers are defined in the `(local_)com4FlowPy.ini` file.

ahuber-bfw (approved):
Looks good from my perspective now!
* The remaining code improvements discussed in #1161 (also using -9999 for non-affected cells in the model domain for all outputs that uses sums) can be addressed at a later stage.

PR #1164, opened 2025-08-06
[com4]: all output rasters have value -9999 in not affected cells

PaulaSp3:
- There is a new parameter in the (local_)com4FlowPyCfg.ini `outputNoDataValue` (default: -9999) that defines the nodata value for the output rasters
-  Then ALL output rasters have the same value in cells taht are not affected by the process (default -9999)
- I've tested in one example, if all -9999 values are at the same locations in every output-raster.
- A further question/ issue?: Should we initialize the rasterArrays consistently? (see a comment below)

### DIFF
--- a/avaframe/runCom4FlowPy.py
+++ b/avaframe/runCom4FlowPy.py
@@ -111,6 +111,7 @@ def main(avalancheDir=''):
         cfgPath["uid"] = uid
         cfgPath["timeString"] = timeString
         cfgPath["outputFiles"] = cfgCustomPaths["outputFiles"]
+        cfgPath["outputNoDataValue"] = cfgCustomPaths["outputNoDataValue"]
 
         com4FlowPy.com4FlowPyMain(cfgPath, cfgSetup)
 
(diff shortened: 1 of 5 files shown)


ahuber-bfw:
Maybe changes are only necessary in `com4FlowPyCfg.ini`, `runCom4FlowPy.py` and `com4FlowPy.py` if we simply provide the cell_counts raster to the  `defineNotAffectedCells()` function in `com4FlowPy.py`?

ahuber-bfw (changes requested):
Looks good and a lot more compact now :)

Maybe we can wait for PR #1173 to be merged and then also resolve issue #1178 in this PR? (adding an .ini option for disabling .tif compression) $\rightarrow$ this would also only require minor changes to `(local)com4FlowPyCfg.ini`, `runCom4FlowPy.py` and `com4FlowPy.py` 

regarding issue #1165 I need to have a closer look still ;) - so maybe could be included in a further PR.

avaframe/runCom4FlowPy.py
```diff
@@ -211,17 +215,17 @@ def readFlowPyinputs(avalancheDir, cfgFlowPy, log):
     avalancheDir = pathlib.Path(avalancheDir)
 
     inputDir = avalancheDir / "Inputs"
-    relFile, available = gI.getAndCheckInputFiles(inputDir, "REL", "Release Area", fileExt="shp")
-    
+    relFile, available, _ = gI.getAndCheckInputFiles(inputDir, "REL", "Release Area", fileExt="shp")
```
ahuber-bfw: added `, _`  wegen Änderungen in den return values von `gI.getAndCheckInputFiles()` ?

fso42:
Standardtest ident

PR #1173, opened 2025-08-18
[com1;in2] Add tiff compression; adjust particles export logic

fso42:
- Particles are only exported if they are in resType or if trackParticles is True

For the compression: 

Each module should implement a useCompression in their cfg IF they want to turn off compression in the case of tiff usage. (done for com1DFA)
By default compression is used!

- [x] Make compression switchable for com4FlowPy via ini -> see #1178


avaframe/com1DFA/com1DFA.py
```diff
@@ -2601,8 +2615,6 @@ def computeEulerTimeStep(
     # particles = updatePosition(cfg, particles, dem, force)
     log.debug("Update position C")
     particles = DFAfunC.updatePositionC(cfg, particles, dem, force, fields, typeStop=0)
-    tCPUPos = time.time() - startTime
-    tCPU["timePos"] = tCPU["timePos"] + tCPUPos
```
awirb: why is this removed?

PR #1219, opened 2025-12-10
[com1;in3] Change time writing [breaking change]

fso42:
Breaking change: by default t=0 is NOT exported anymore

feat(com1DFA): refine tSteps handling for timestep exports
- Export initial timestep (t=0) only if explicitly specified in `tSteps`.
- Updated default behavior to export only final timestep when `tSteps` is empty.

  1. avaframe/com1DFA/com1DFACfg.ini:16-20
    - Changed default from tSteps = 1 to tSteps = (empty)
    - Updated comments to reflect new behavior
  2. avaframe/in3Utils/fileHandlerUtils.py:387-389
    - Removed filtering that excluded t=0 from dtSave array
  3. avaframe/com1DFA/com1DFA.py:2097
    - Made initial timestep export conditional: only exports if t=0 is in dtSave
  4. avaframe/com1DFA/com1DFA.py:2131-2134
    - Made Tsave initialization conditional based on whether t=0 is exported
  5. docs/moduleCom1DFA.rst:258
    - Updated documentation: "initial time step" → "final time step" to reflect new default

### DIFF
--- a/avaframe/com1DFA/com1DFA.py
+++ b/avaframe/com1DFA/com1DFA.py
@@ -2103,15 +2103,21 @@ def DFAIterate(cfg, particles, fields, dem, inputSimLines, outDir, cuSimName, si
         outDirData = outDir / "particles"
         fU.makeADir(outDirData)
 
-    # export initial time step only if t=0 is explicitly in dtSave
-    if cfg["EXPORTS"].getboolean("exportData") and (dtSave.size > 0 and np.any(dtSave <= 1.0e-8)):
+    # Save original dtSave for initial timestep decisions
+    dtSaveOriginal = dtSave.copy()
+
+    # export initial time step only if t=0 is explicitly in dtSaveOriginal
+    if cfg["EXPORTS"].getboolean("exportData") and (
+        dtSaveOriginal.size > 0 and np.any(dtSaveOriginal <= 1.0e-8)
+    ):
         exportFields(cfg, t, fields, dem, outDir, cuSimName, TSave="initial")
 
         if "particles" in resTypes:
             savePartToPickle(particles, outDirData, cuSimName)
 
         # Update dtSave to remove the initial timestep we just saved
-        dtSave = updateSavingTimeStep(dtSave, cfgGen, t)
+        dtSave = updateSavingTimeStep(dtSaveOriginal, cfgGen, t)
+
 
     # export particles properties for visulation
     if cfg["VISUALISATION"].getboolean("writePartToCSV"):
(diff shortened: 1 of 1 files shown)


avaframe/com1DFA/com1DFA.py
```diff
@@ -2093,15 +2097,21 @@ def DFAIterate(cfg, particles, fields, dem, inputSimLines, outDir, cuSimName, si
     t = particles["t"]
     log.debug("Saving results for time step t = %f s", t)
 
-    # export initial time step
-    if cfg["EXPORTS"].getboolean("exportData"):
+    # Initialize particles output directory if needed
+    if "particles" in resTypes:
+        outDirData = outDir / "particles"
+        fU.makeADir(outDirData)
```
awirb: why outside of exportData if?

avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -546,8 +546,6 @@ thFromIni =
 
 
 [REPORT]
-# which result parameters shall be included as plots in report - separated by |
-plotFields = ppr|pft|pfv
 # units for output variables
 unitppr = kPa
 unitpft = m
```
awirb: is this still needed?

avaframe/com1DFA/com1DFA.py
```diff
@@ -1658,8 +1658,11 @@ def initializeFields(cfg, dem, particles, releaseLine):
     cfgGen = cfg["GENERAL"]
     # what result types are desired as output (we need this to decide which fields we actually need to compute)
     resTypes = fU.splitIniValueToArraySteps(cfgGen["resType"])
-    resTypesReport = fU.splitIniValueToArraySteps(cfg["REPORT"]["plotFields"])
-    resTypesLast = list(set(resTypes + resTypesReport))
+    # ensure at least one field type is present for internal computations
+    # if resTypes only contains particles add pfv
+    validFieldTypes = [rt for rt in resTypes if rt not in ["particles"]]
+    if len(validFieldTypes) == 0:
+        resTypes.append("pfv")
```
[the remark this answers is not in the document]
fso42: If only particles exist in the resType, quite a few of our routines fail (eg. max routines). Therefore particles are removed and handled extra, but to not have an empyt resType dict (also leads to problem), pfv is added

avaframe/com1DFA/com1DFA.py
fso42: Because the if below only covers the initial timestep export. If that one is not hit, the next (intermediate and final) would need the  same check. To keep it simple, I added it here.

avaframe/com1DFA/com1DFACfg.ini
fso42: Nope, removing it

PR #1258, opened 2026-03-25
[runScripts]: wrap script execution in __main__ block; fixes #1247

### LINKED ISSUE #1247: [tests] Problems when running standard tests under Windows
**Bug**
When running `runStandardTestsCom1DFA.py` and `runUpdateBenchmarkTestsCom1DFA` on master branch and with default settings under Windows, following error message occurs:

Image

ChatGPT assumes the following reason:

On Windows, multiprocessing uses the spawn method by default (not fork as on Linux).
This means:

- Every new subprocess instance re-imports your main script.
- If your code starts processes directly during import (e.g., Manager(), Process(), Pool(), etc.), an infinite loop is created.
- That is why Windows requires protecting the entry point with:

if __name__ == "__main__":

The error needs to be confirmed.

**Desktop**
 - OS: Windows 11 Pro
 - Version 25H2

fso42:
Wraps the top-level execution code in `runStandardTestsCom1DFA.py` and
`runUpdateBenchmarkTestsCom1DFA.py` in `if __name__ == "__main__"` guards.

This prevents unintended execution when these scripts are imported as
modules (e.g. during test collection or parallel processing forks),
which was the root cause of #1247.

### DIFF
--- a/avaframe/ana3AIMEC/aimecTools.py
+++ b/avaframe/ana3AIMEC/aimecTools.py
@@ -1219,7 +1219,7 @@ def computeRunOut(
     resAnalysisDF.loc[simRowHash, "zRelease"] = zThalweg[cUpper]
     resAnalysisDF.loc[simRowHash, "elevRel"] = zThalweg[cUpper]
     resAnalysisDF.loc[simRowHash, "deltaZ"] = zThalweg[cUpper] - zThalweg[cLower]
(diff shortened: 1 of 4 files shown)


fso42:
Standardtests ident

JuLa96 (approved):
The scripts are running under Windows now.

fso42:
Standardtests ok, apart from known snowslide test

PR #1259, opened 2026-04-07
[com5]: set detrainment to False 

awirb:


fso42 (approved):

fso42:
Standardtest are ident, APART from avaHof (expected) -> needs benchmark updating

PR #1263, opened 2026-04-22
Update Scarp.py

dwolfsch:
An Update for the scarp script, so that it does not allow negative values in the generated raster files anymore.

### DIFF
--- a/avaframe/com6RockAvalanche/scarp.py
+++ b/avaframe/com6RockAvalanche/scarp.py
@@ -149,8 +149,14 @@ def scarpAnalysisMain(cfg, baseDir):
         )
     else:
         raise ValueError("Unsupported method. Choose 'plane' or 'ellipsoid'.")
+        
 
     hRelData = dem["rasterData"] - scarpData
+    min_hrel = hRelData.min()
+    if min_hrel < 0:
+        log.warning("hRel: most negative clipped value = %.2f m", min_hrel)
+    hRelData = np.maximum(0, hRelData)   #enforce non-negative release
+
     
     #Compute and log excavated volume
     cellArea = abs(dem["header"]["cellsize"] ** 2)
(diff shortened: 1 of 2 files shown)


avaframe/com6RockAvalanche/scarp.py
```diff
@@ -373,6 +388,15 @@ def calculateScarpWithEllipsoids(elevData, periData, elevTransform, ellipsoids):
...
 
-            scarpData[row, col] = scarpVal if periData[row, col] > 0 else elevData[row, col]
+            if periData[row, col] > 0:
+                if scarpVal < 0:
+                    min_clipped = min(min_clipped, scarpVal)
+                scarpData[row, col] = max(0, scarpVal)
+            else:
+                scarpData[row, col] = elevData[row, col]
+
+    # nach der Schleife, vor return:
```
fso42, line 398: German comment, either translate or remove

avaframe/com6RockAvalanche/scarp.py
```diff
@@ -373,6 +388,15 @@ def calculateScarpWithEllipsoids(elevData, periData, elevTransform, ellipsoids):
                     totalDepth = baseDepth + tiltEffect + z0
                     scarpVal = min(scarpVal, elevData[row, col] - totalDepth)
 
-            scarpData[row, col] = scarpVal if periData[row, col] > 0 else elevData[row, col]
+            if periData[row, col] > 0:
+                if scarpVal < 0:
```
fso42: This is inconsistent with the plane method: the plane method checks val = min(elevData, scarpVal) and then checks val < 0. The ellipsoid method skips the min(elevData, scarpVal) step entirely before the check, so the guard condition is inconsistent.
Is this wanted?

fso42 (changes requested):

fso42 (approved):

fso42:
Standardtest ok, apart from timedep and snow slide (expected)

PR #1273, opened 2026-05-11
[com1]: add option to only save certain particle props

awirb:


avaframe/com1DFA/com1DFACfg.ini
```diff
@@ -570,4 +570,7 @@ exportData = True
 exportRasters = False
 # use LZW compression when writing TIFF raster files
 useCompression = True
+# particle properties - list all properties that shall be saved, t is always added
```
fso42, line 573: Add info that particle property list can be found in the documentation

avaframe/com1DFA/com1DFA.py
```diff
@@ -2846,16 +2849,32 @@ def savePartToPickle(dictList, outDir, logName):
...
+    # create list of particle properties and append t (time info)
+    if particleProperties != "":
+        particleProperties = particleProperties.split("|")
+        if "t" not in particleProperties:
+            particleProperties.append("t")
+
     if isinstance(dictList, list):
         for dict in dictList:
+            if particleProperties != "":
+                particlesToSave = {key: dict[key] for key in particleProperties}
```
fso42: Here (and further down): if key is not available (eg. typo) it just fails with KeyError.  Maybe validate the particle properties first in this function and throw an error right away with info about the wrong key.

avaframe/com1DFA/com1DFA.py
```diff
@@ -2846,16 +2849,32 @@ def savePartToPickle(dictList, outDir, logName):
...
+            pickle.dump(particlesToSave, fi)
             fi.close()
     else:
+        if particleProperties != "":
+            particlesToSave = {key: dictList[key] for key in particleProperties}
+        else:
+            particlesToSave = dictList
         fi = open(outDir / ("particles_%s_%09.4f.pickle" % (logName, dictList["t"])), "wb")
-        pickle.dump(dictList, fi)
+        pickle.dump(particlesToSave, fi)
```
fso42: This and the above could be written in only one statement after the if/else block

fso42 (approved):

avaframe/com1DFA/com1DFACfg.ini
fso42, line 573: On second go around: why add this to [EXPORT] and not to [TRACKPARTICLES] ? As it needs trackParticles to be true?

avaframe/com1DFA/com1DFA.py
```diff
@@ -2846,16 +2845,107 @@ def savePartToPickle(dictList, outDir, logName):
...
+        "dmEnt",
+        "stoppedParticles",
+        "massInitialized",
+        "massEntrained",
+        "massDetrained",
+        "massStopped",
+    ]
+
+    # create list of particle properties and append t (time info)
+    if isinstance(cfg, configparser.ConfigParser):
```
fso42: If trackParticle is True, particleProperties stays empty (by default) and exportParticleProperties contains something, it fails with an KeyError.  as the split adds an empty ''.

avaframe/com1DFA/com1DFA.py
```diff
@@ -2834,7 +2833,7 @@ def releaseSecRelArea(cfg, particles, fields, dem, zPartArray0, reportAreaInfo):
     return particles, zPartArray0, reportAreaInfo
 
 
-def savePartToPickle(dictList, outDir, logName):
+def savePartToPickle(dictList, outDir, logName, cfg=""):
```
fso42, line 2871: The whole function is very complicated to read, I'll make an extra commit with a reworked suggestion.

fso42 (changes requested):

avaframe/com1DFA/com1DFACfg.ini
awirb: no trackParticles is a 'separate' functionality -particles can also be saved if tracking specific particles is not active

fso42 (approved):

fso42:
Standardtests ident, apart from known com5SSL

PR #1274, opened 2026-05-13
[com5] Fix com5 bug on windows plus inconsistencies in code

fso42:
Reported by CT (com5 hang on althof testcase). During analysis other problems cropped up, which are fixed now                                                  
                                                                                                                                                                  
   ## Summary                                                                                                                                                     
                                                                                                                                                                  
   Fixes several bugs in com1DFA discovered during simulations: circular reference in stopped particles, stale bond indices after particle removal, out-of-bounds 
 heap corruption from mismatched test array sizes, plus hardening against NaN/div-by-zero in reprojection.                                                        
                                                                                                                                                                  
   ## Changes                                                                                                                                                     
                                                                                                                                                                  
   ### 1. Remove circular reference in `stoppedParticles` (9f83630f)                                                                                              
   `particles["stoppedParticles"] = particles` created a self-referencing dict that overwrote the correct stopped-particle arrays (x, y, m, ID) with full         
 flowing-particle data. Downstream in `updateFieldsC` this caused stopped mass rasters to use flowing instead of stopped masses. It also prevented garbage        
 collection and broke serialization. The line was removed; `stoppedParticles` is already correctly populated earlier.                                             
                                                                                                                                                                  
   ### 2. Remap bond indices after particle removal (e425c76d)                                                                                                    
   `removedBonds` now builds an old-to-new particle index mapping and remaps both `bondStart` (reindexed by new particle indices) and `bondPart` (target indices  
 remapped). `removePart` skips bond keys since they are already rebuilt. Previously `bondPart` stored original indices that became stale after `np.extract`       
 compacted arrays, causing silent data corruption or out-of-bounds access.                                                                                        
                                                                                                                                                                  
   ### 3. Fix division-by-zero in `distConservProjectionIteratrive` (f47629a1)                                                                                    
   Added `distn == 0` guard before `dist / distn` to prevent NaN for stopped particles.                                                                           
                                                                                                                                                                  
   ### 4. Ensure arrays returned by `updateFieldsC` are owned copies (ab231116)                                                                                   
   Multiple `fields` and `particles` assignments now use `.copy()` when converting to `np.asarray`, preventing issues with memory views that could lead to        
 dangling references or double-free (BUG COM5). Adjusted `bondDist` handling to avoid problematic views during inplace modification. Added tests verifying        
 `owndata=True` and `base is None`.

### DIFF
--- a/avaframe/com1DFA/DFAToolsCython.pyx
+++ b/avaframe/com1DFA/DFAToolsCython.pyx
@@ -675,6 +675,9 @@ cpdef (double, double, double, int, int, int, double, double, double, double) di
     # second step: conserve the distance dist (correction ov the position)
     # measure distance between this new point and the previous time step position
     distn = norm(xNew-xPrev, yNew-yPrev, zNew-zPrev)
+    # guard against stopped particle (distn == 0)
+    if distn == 0:
+      break
     # adjust position on the Xprev, Xnew line
     xNew = xPrev + (xNew-xPrev) * dist / distn
     yNew = yPrev + (yNew-yPrev) * dist / distn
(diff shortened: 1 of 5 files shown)


fso42:
Standardtests ident, apart from known com5 com1(timedep)

PR #1280, opened 2026-05-26
[in2]: add function to convert raster to nc file

PaulaSp3:
The nc files have name convention (including the timesteps) to represent time steps in ParaView.


avaframe/in2Trans/rasterUtils.py
```diff
@@ -227,3 +231,57 @@ def getRasterFileTypeFromHeader(header):
         raise AssertionError("Unsupported driver for DEM %s" % header["driver"])
 
     return fileType
+
+
+def convertRasterToNcFile(inputDir, infileExt, infileSuffix, outputDir, outFileName="", crs=None):
```
fso42, line 240: If crs is actually none, i.e. no crs is given, how does gdal handle this? Please check. Maybe the a_srs flag should only be used when crs is not None?

avaframe/in2Trans/rasterUtils.py
```diff
@@ -227,3 +231,57 @@ def getRasterFileTypeFromHeader(header):
...
+    )
+
+    for i, filename in enumerate(inputFiles):
+        if outFileName == "":
+            cleanName = re.sub(r"_t[\d.]+", "", filename.stem)
+            outputFileName = f"{cleanName}_{i:04d}.nc"
+        else:
+            outputFileName = f"{outFileName}_{i:04d}.nc"
+
+        subprocess.run(
```
fso42: Mh, maybe it would be helpfull to check whether the subprocess actually worked? with check=True or similar. Right now, if gdal fails, there is no feedback

avaframe/in2Trans/rasterUtils.py
```diff
@@ -227,3 +231,57 @@ def getRasterFileTypeFromHeader(header):
...
+    inputDir = pathlib.Path(inputDir)
+    outputDir = pathlib.Path(outputDir)
+
+    os.makedirs(outputDir, exist_ok=True)
+
+    def extractTime(filename):
+        match = re.search(r"_t([\d.]+)\.", filename.name)
+        return float(match.group(1)) if match else float("inf")
+
+    inputFiles = list([file for file in inputDir.glob(f"*{infileSuffix}*.{infileExt}")])
```
fso42: The list is not necessary, as [file for file...] already produces a list, right?

avaframe/in2Trans/rasterUtils.py
```diff
@@ -227,3 +231,57 @@ def getRasterFileTypeFromHeader(header):
...
+
+    inputFiles.sort(
+        key=extractTime,
+    )
+
+    for i, filename in enumerate(inputFiles):
```
fso42, line 274: Mh, to be able to debug, maybe add a log info about which file are being converted

avaframe/in2Trans/rasterUtils.py
```diff
@@ -227,3 +231,57 @@ def getRasterFileTypeFromHeader(header):
...
+        path to folder where netCDF files will be stored
+    outFileName: str
+        name of output netCDF file (if "", the same name of input raster file is used)
+    crs: str
+        overwrite crs
+    """
+    inputDir = pathlib.Path(inputDir)
+    outputDir = pathlib.Path(outputDir)
+
+    os.makedirs(outputDir, exist_ok=True)
```
fso42: Use pathlib.Path.mkdir instead of os.makedirs, this import already exists. so you can remove the import os too...

fso42 (changes requested):

fso42 (approved):

avaframe/in2Trans/rasterUtils.py
```diff
@@ -227,3 +230,81 @@ def getRasterFileTypeFromHeader(header):
...
+        overwrite crs
+    """
+    inputDir = pathlib.Path(inputDir)
+    outputDir = pathlib.Path(outputDir)
+
+    outputDir.mkdir(exist_ok=True)
+
+    def extractTime(filename):
+        match = re.search(r"_t([\d.]+)\.", filename.name)
+        return float(match.group(1)) if match else float("inf")
```
awirb: consider moving this outside of this function as def _extractTime

avaframe/in2Trans/rasterUtils.py
```diff
@@ -227,3 +230,81 @@ def getRasterFileTypeFromHeader(header):
...
+                    check=True,
+                    capture_output=True,
+                    text=True,
+                )
+            log.info("%s is converted into %s and saved in %s" % (filename, outputFileName, outputDir))
+        except subprocess.CalledProcessError as e:
+            message = "%s could not be converted: %s." % (filename, e.stderr)
+            log.info(message)
+
+    log.info("%s raster files are converted and saved in %s" % (len(inputFiles), outputDir))
```
awirb: is this required if the is the log.info in line 305?

PaulaSp3: I suggest yes, especially when converting many files.

awirb: will this log files as converted also if the except occurred?

avaframe/in2Trans/rasterUtils.py
```diff
@@ -227,3 +230,81 @@ def getRasterFileTypeFromHeader(header):
...
+            outputFileName = f"{outFileName}_{i:04d}.nc"
+
+        try:
+            if crs is None:
+                subprocess.run(
+                    [
+                        "gdal_translate",
+                        "-of",
+                        "netCDF",
+                        inputDir / filename,
```
awirb:
```suggestion
                        inputDir / filename.name,
```

this is required otherwise you add the full file path again

avaframe/in2Trans/rasterUtils.py
```diff
@@ -227,3 +230,81 @@ def getRasterFileTypeFromHeader(header):
...
+                )
+            else:
+                subprocess.run(
+                    [
+                        "gdal_translate",
+                        "-of",
+                        "netCDF",
+                        "-a_srs",
+                        crs,
+                        inputDir / filename,
```
awirb:
```suggestion
                        inputDir / filename.name,
```

same here

avaframe/in2Trans/rasterUtils.py
```diff
@@ -227,3 +230,81 @@ def getRasterFileTypeFromHeader(header):
         raise AssertionError("Unsupported driver for DEM %s" % header["driver"])
 
     return fileType
+
+
+def convertRasterToNcFile(inputDir, infileExt, infileSuffix, outputDir, outFileName="", crs=None):
```
awirb, line 240: if infileSuffix is "" empty string it doesn't work - however could be a standard use case to just convert all files within a folder and given file extension, consider allowing for this

awirb (changes requested):

fso42:
Standardtests ident, apart from know com5

PR #1281, opened 2026-05-27
[com9] Add com9 compilation (instead of fixed .exe)

fso42:
## Summary

Remove precompiled MoT-Voellmy binaries from version control and instead compile them at package time. A build script fetches the C source from the upstream GitHub repository and compiles for the target platform.

## Changes

- **New build script** (`avaframe/com9MoTVoellmy/_buildMoTVoellmy.py`): discovers the latest C source from the upstream repo via the GitHub API, caches it, and compiles with `gcc`
- **Pixi task**: `pixi run compilemot` for manual compilation, optionally with a local `.c` file path
- **setup.py hook**: compiles the binary automatically during `pip install` (best-effort, warns if gcc is missing)
- **CI wheels**: `CIBW_BEFORE_BUILD_*` runs the build script in each platform environment
- **CI sdist**: compiles Linux natively and cross-compiles Windows via `mingw-w64`, ships both binaries in the source tarball


awirb (approved):

fso42:
Standardtests ident, apart from know com5

PR #1290, opened 2026-06-08
[com1]: add saving option for resistance fields

awirb:


avaframe/com1DFA/com1DFA.py
```diff
@@ -2206,6 +2208,18 @@ def DFAIterate(cfg, particles, fields, dem, inputSimLines, outDir, cuSimName, si
     particles["dt"] = dt
     t = t + dt
 
+    # save initial resistance cRes field
+    if (cfgGen["simTypeActual"] in ["entres", "res"]) and cfg["EXPORTS"].getboolean("exportRasters"):
+        outDirRes = outDir / "internalRasters"
```
fso42: outDirRes is referenced inside another if. This works now, since both ifs have the same conditionals, but if one changes, this might lead to problems.  Put this outside of the first if...

avaframe/com1DFA/com1DFA.py
```diff
@@ -2448,6 +2462,16 @@ def DFAIterate(cfg, particles, fields, dem, inputSimLines, outDir, cuSimName, si
             % particles["nExitedParticles"]
         )
 
+    # save final cRes raster
+    if (cfgGen["simTypeActual"] in ["entres", "res"]) and cfg["EXPORTS"].getboolean("exportRasters"):
+        IOf.writeResultToRaster(
+            dem["originalHeader"],
+            fields["cResRasterOrig"],
```
fso42: Why is cResRasterOrig and not cResRaster saved here? This is identical with what is being saved at the inital time step, right?

avaframe/com1DFA/com1DFA.py
```diff
@@ -2206,6 +2208,18 @@ def DFAIterate(cfg, particles, fields, dem, inputSimLines, outDir, cuSimName, si
     particles["dt"] = dt
     t = t + dt
 
+    # save initial resistance cRes field
+    if (cfgGen["simTypeActual"] in ["entres", "res"]) and cfg["EXPORTS"].getboolean("exportRasters"):
```
fso42: add a saving log similar to the other raster exports (eg. line 1238)

avaframe/com1DFA/com1DFA.py
```diff
@@ -2448,6 +2462,16 @@ def DFAIterate(cfg, particles, fields, dem, inputSimLines, outDir, cuSimName, si
             % particles["nExitedParticles"]
         )
 
+    # save final cRes raster
+    if (cfgGen["simTypeActual"] in ["entres", "res"]) and cfg["EXPORTS"].getboolean("exportRasters"):
```
fso42, line 2473: also a log info about saving file

fso42 (changes requested):

avaframe/com1DFA/com1DFA.py
awirb: moved the outDirRes creation outside the if - so it will be created even if not needed

avaframe/com1DFA/com1DFA.py
awirb: the cResRasterOrig is overwritten in each time step and the one at the final time steps carries the information of when cells of the resistance area have been removed if thresholds were exceeded so it is not the same as the initial one - see line 403 in com1DFATools.py

fso42 (approved):

fso42:
Standardtests ident (apart from known com5)

PR #1296, opened 2026-06-16
[ana5] Add extBottomOption to extend the DFA path to the deposit front

NUForever:
## What this adds

A new option `extBottomOption` in `DFAPathGeneration` controlling how the
mass-averaged path is extended at the bottom:

- `extBottomOption = 0` (default, unchanged): the current fixed-length
  extrapolation (`extendProfileBottom`), which extends the path in the
  direction of its last points by `factBottomExt * sMax`, clipped at the
  DEM border.
- `extBottomOption = 1` (new): extend the path to the front of the
  simulated deposit.

## Motivation

The mass-averaged path ends short of the deposit once the front
decelerates. The default bottom extension extrapolates a straight line of
fixed relative length, independently of where the avalanche actually
stopped, so the path end does not coincide with the simulated runout.
Targeting the deposit front yields a path that ends at the simulated
runout, which is useful for corridor-scale and road-crossing analyses.

## How it works

1. The front of the deposit is located in the peak flow thickness field
   (`pft`) as the flow-thickness-weighted centroid of the flow cells in the
   lowest part of the flow elevation range (`findFlowFront`), with a
   lowest-cell fallback for flat deposits and a snap to the deposit when the
   centroid falls between disjoint lobes.
2. The path end is connected to that front by a Dijkstra least-cost path
   over the DEM (`leastCostPath`). The cost of a step is its horizontal
   length plus a penalty on the positive elevation gain and on the distance
   from the flow footprint, so the extension descends along the deposit.
   Distances use meters (`distance_transform_edt` with `sampling=cellsize`)
   so the penalties are resolution independent.
3. If the front cannot be located or reached, or no flow thickness field is
   available, the path falls back to the fixed-length extension, so the
   AIMEC buffer is preserved.

The option is dispatched in `extendDFAPath` (which gains an optional
`fieldFT` argument, mirroring the existing `extTopOption`); the `pft` field
is read in `generatePathAndSplitpoint` only when the option is active. New
parameters live in `DFAPathGenerationCfg.ini` (`ftThreshold`,
`lowFrontFraction`, `upSlopePenalty`, `flowDistPenalty`) and `resType`
gains `pft`.

## Default behaviour

Unchanged. With `extBottomOption = 0` (the default) the existing pipeline
is not affected.

## Tests

Unit tests added for `findFlowFront` (front band, flat deposit, two-lobe
snap, no flow), `leastCostPath` (channel routing, unreachable goal behind
nodata), `extendProfileToFront` (extension to the front, both fallbacks)
and `readPeakFT` (lookup by name and by hash, missing field). Docs updated
in `docs/moduleAna5Utils.rst`.

## Context

Developed for the corridor-scale thalweg extraction on Route 115-CH (Chile)
and validated there against 208 manually delineated reference thalwegs
(median endpoint distance 64 m).

### DIFF
--- a/avaframe/ana5Utils/DFAPathGeneration.py
+++ b/avaframe/ana5Utils/DFAPathGeneration.py
@@ -96,9 +96,9 @@ def generatePathAndSplitpoint(avalancheDir, cfgDFAPath, cfgMain, runDFAModule):
         pathFromPart = cfgDFAPath['PATH'].getboolean('pathFromPart')
         resampleDistance = cfgDFAPath['PATH'].getfloat('nCellsResample') * dem['header']['cellsize']
         # peak flow thickness field, only needed when extending the path to the deposit front
-        fieldFT = None
+        fieldPFT = None
         if extendToFront:
-            fieldFT = readPeakFT(peakFilesDF, simName)
+            fieldPFT = readPeakFT(peakFilesDF, simName)
         # get the mass average path
         avaProfileMass, particlesIni = generateMassAveragePath(avalancheDir, pathFromPart, simName, dem,
                                                                addVelocityInfo=cfgDFAPath['PATH'].getboolean('addVelocityInfo'))
@@ -109,7 +109,7 @@ def generatePathAndSplitpoint(avalancheDir, cfgDFAPath, cfgMain, runDFAModule):
         parabolicFit = getParabolicFit(cfgDFAPath['PATH'], avaProfileMass, dem)
         # here the avaProfileMass given in input is overwritten and returns only an x, y, z extended profile
         avaProfileMass = extendDFAPath(cfgDFAPath['PATH'], avaProfileMass, dem, particlesIni,
-                                       fieldFT=fieldFT)
+                                       fieldPFT=fieldPFT)
         # resample path and keep track of start and end of mass averaged part
         avaProfileMass = resamplePath(cfgDFAPath['PATH'], dem, avaProfileMass)
         # get split point
@@ -334,7 +334,7 @@ def readPeakFT(peakFilesDF, simName):
 
     Returns
     --------
-    fieldFT: numpy array
+    fieldPFT: numpy array
         peak flow thickness raster (same grid as the simulation dem),
         None if no pft peak field is available for this simulation
     """
@@ -348,7 +348,7 @@ def readPeakFT(peakFilesDF, simName):
     return IOf.readRaster(peakFilesDF.loc[index[0], 'files'])['rasterData']
 
 
-def extendDFAPath(cfg, avaProfile, dem, particlesIni, fieldFT=None):
+def extendDFAPath(cfg, avaProfile, dem, particlesIni, fieldPFT=None):
     """ extend the DFA path at the top and bottom
     avaProfile with x, y, z, s information
 
@@ -358,8 +358,8 @@ def extendDFAPath(cfg, avaProfile, dem, particlesIni, fieldFT=None):
         configuration object with:
 
         - extTopOption: int, how to extend towards the top? 0 for heighst point method, a for largest runout method
-        - extBottomOption: int, how to extend towards the bottom? 0 for the fixed-length extrapolation
-        method, 1 for the least-cost extension to the deposit front (requires fieldFT)
+        - extBottomOption: int, how to extend towards the bottom? 0 for the straight-line extrapolation
+        method, 1 for the least-cost extension to the deposit front (requires fieldPFT)
         - nCellsResample: int, resampling length is given by nCellsResample*demCellSize
         - nCellsMinExtend: int, when extending towards the bottom, take points at more
         than nCellsMinExtend*demCellSize from last point to get the direction
@@ -373,7 +373,7 @@ def extendDFAPath(cfg, avaProfile, dem, particlesIni, fieldFT=None):
         dem dict
     particlesIni: dict
         initial particles dict
-    fieldFT: numpy array, optional
+    fieldPFT: numpy array, optional
         peak flow thickness field on the dem grid, only used if extBottomOption = 1
 
     Returns
@@ -386,12 +386,12 @@ def extendDFAPath(cfg, avaProfile, dem, particlesIni, fieldFT=None):
     avaProfile, _ = gT.prepareLine(dem, avaProfile, distance=resampleDistance, Point=None)
     avaProfile = extendProfileTop(cfg.getint('extTopOption'), particlesIni, avaProfile)
     if cfg.getint('extBottomOption', fallback=0) == 1:
-        if fieldFT is None:
-            log.warning('extBottomOption is 1 but no flow thickness field was provided, '
-                        'falling back to the fixed-length bottom extension')
+        if fieldPFT is None:
+            log.warning('extBottomOption is 1 but no peak flow thickness field was provided, '
+                        'falling back to the straight-line bottom extension')
             avaProfile = extendProfileBottom(cfg, dem, avaProfile)
         else:
-            avaProfile = extendProfileToFront(cfg, dem, avaProfile, fieldFT)
+            avaProfile = extendProfileToFront(cfg, dem, avaProfile, fieldPFT)
     else:
         avaProfile = extendProfileBottom(cfg, dem, avaProfile)
     return avaProfile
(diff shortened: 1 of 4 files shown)


avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -84,6 +87,10 @@ def generatePathAndSplitpoint(avalancheDir, cfgDFAPath, cfgMain, runDFAModule):
         log.info('Computing avalanche path from simulation: %s', simName)
         pathFromPart = cfgDFAPath['PATH'].getboolean('pathFromPart')
         resampleDistance = cfgDFAPath['PATH'].getfloat('nCellsResample') * dem['header']['cellsize']
+        # peak flow thickness field, only needed when extending the path to the deposit front
+        fieldFT = None
+        if cfgDFAPath['PATH'].getint('extBottomOption', fallback=0) == 1:
+            fieldFT = readPeakFT(avalancheDir, simName)
```
fso42: In your readPeakFT function, makeSimDF(inputDir, …) is used, which parses all peak files. Since we are inside the simDF.loop, this happens for each sim and is just a re-reading and re-parsing. The data frame could be build before the loop and passed as argument, if possible.

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -523,6 +573,203 @@ def extendProfileBottom(cfg, dem, profile):
...
+        cell of the front, (None, None) if there is no flow above ftThreshold
+    """
+    flow = (fieldFT > ftThreshold) & np.isfinite(demRaster)
+    if not flow.any():
+        return None, None
+    rows, cols = np.nonzero(flow)
+    elev = demRaster[rows, cols]
+    lowBand = elev <= elev.min() + lowFrontFraction * (elev.max() - elev.min())
+    weight = fieldFT[rows, cols] * lowBand
+    frontRow = int(round(np.sum(rows * weight) / weight.sum()))
```
fso42, line 695: If I understood this correctly, then weight.sum() is never zero, correct? Otherweise a divide-by-zero guard would be necessary. If my understanding is correct, then it is not needed, but maybe a quick comment would help future maintainers

fso42 (changes requested):
Thank  you Ivan for your PR. I have 2 comments, one is only minor.

avaframe/ana5Utils/DFAPathGeneration.py
NUForever: Good catch, fixed in eb692ff8. makeSimDF is now called once before the simulation loop in generatePathAndSplitpoint and the resulting dataframe is passed to readPeakFT, so the peak files are parsed a single time for the whole run instead of once per simulation.

avaframe/ana5Utils/DFAPathGeneration.py
NUForever: You understood it correctly, weight.sum() is always positive: the lowest flow cell is always in lowBand and, being a flow cell, has fieldFT > ftThreshold >= 0, so its weight is strictly positive. I added a comment to that effect in eb692ff8 instead of a guard.

fso42 (approved):

awirb:
Does the new extension need peak flow thickness  or is also just flow thickness of the final time step sufficient? the fieldFT variable is initialized with the peak flow thickness and pft has also been added as a required simulation result but later on when used in docstring also referred to as flow thickness field? I guess the peak flow thickness field is needed as otherwise the computation of the lowBand (lowest part of the flow elevation range) is changed

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -80,10 +83,22 @@ def generatePathAndSplitpoint(avalancheDir, cfgDFAPath, cfgMain, runDFAModule):
...
+    if extendToFront:
+        peakFilesDir = pathlib.Path(avalancheDir, 'Outputs', 'com1DFA', 'peakFiles')
+        peakFilesDF = fU.makeSimDF(peakFilesDir, avaDir=avalancheDir)
+
     for simName, simDFrow in simDF.iterrows():
         log.info('Computing avalanche path from simulation: %s', simName)
         pathFromPart = cfgDFAPath['PATH'].getboolean('pathFromPart')
         resampleDistance = cfgDFAPath['PATH'].getfloat('nCellsResample') * dem['header']['cellsize']
+        # peak flow thickness field, only needed when extending the path to the deposit front
+        fieldFT = None
```
awirb: if it needs to be pft and not ft of final time step, consider renaming to fieldPFT

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -316,6 +358,8 @@ def extendDFAPath(cfg, avaProfile, dem, particlesIni):
         configuration object with:
 
         - extTopOption: int, how to extend towards the top? 0 for heighst point method, a for largest runout method
+        - extBottomOption: int, how to extend towards the bottom? 0 for the fixed-length extrapolation
```
awirb: I think fixed-length is a bit misleading, as this is a setting that can be chosen, maybe calling it a straight-line extrapolation is a bit more descriptive of that choice?

docs/moduleAna5Utils.rst
```diff
@@ -119,10 +119,24 @@ There are two options available to extend the mass-averaged path profile in the
...
+We also extend the path at the bottom, to have some buffer in the runout area. Two options exist
+(``extBottomOption``):
+
+* ``extBottomOption = 0`` (default): find the direction of
+  the path given by the last few points within the path in the x,y domain (all points at a distance ``nCellsMinExtend`` *
+  cellSize < distance < ``nCellsMaxExtend`` * cellSize)) and extend in this direction by a given factor
+  (``factBottomExt``) of the total length of the path :math:`s`.
+
+* ``extBottomOption = 1``: extend the path to the front of the deposit. Since the mass-averaged path
+  ends short of the deposit once the front decelerates, this option first locates the front as the
```
awirb: by definition the mass averaged path ends at the last (time step wise) center of mass point - so not at the furthest reaching cell

docs/moduleAna5Utils.rst
```diff
@@ -119,10 +119,24 @@ There are two options available to extend the mass-averaged path profile in the
...
+  flow-thickness-weighted centroid of the flow cells lying in the lowest ``lowFrontFraction`` of the
```
awirb: so the parameter lowFrontFraction exerts a strong control on how close to the 'actual end of the deposit' this point is located, consider adding information on the choice of this parameter - default it is set to 0.05 right?

awirb (changes requested):

NUForever:
Thanks for the review, Anna. To your question: it is the peak flow thickness (pft). The flow thickness of the final time step could be used to locate the front as well, but I went with pft because it is a default com1DFA output that is always computed, independent of whether the path is built from the particles or from the fields, while the final time step thickness would depend on that choice. The pft also covers the full extent the flow reached, which is what we need to place the front. I also renamed nrows/ncols to nRows/nCols in leastCostPath for camelCase consistency, as you suggested. All points are addressed in 331c5439, with replies in the inline threads.

avaframe/ana5Utils/DFAPathGeneration.py
NUForever: Renamed fieldFT to fieldPFT and made the docstrings say peak flow thickness (pft) consistently in 331c5439. The reasoning for pft over the final time step thickness is in the conversation comment.

avaframe/ana5Utils/DFAPathGeneration.py
NUForever: Agreed, renamed to straight-line extrapolation in the docstrings, log messages and the docs (331c5439).

docs/moduleAna5Utils.rst
NUForever: Corrected, the docs now say the path ends at the last center of mass position, which lies behind the deposit front (331c5439).

docs/moduleAna5Utils.rst
NUForever: Added, a smaller lowFrontFraction places the front closer to the furthest reaching deposit cells, and the default (0.05) is now documented (331c5439).

awirb (approved):

fso42:
Standardtests ok (apart from known com5)

PR #1299, opened 2026-06-23
[log]: add collection and summary of comparison warnings

fso42:
- Introduced `collectComparisonWarnings` to identify and structure flagged differences between simulated, benchmark, and configuration data.

Add a per-test comparison-warnings summary to runStandardTestsCom1DFA.py, surfacing the differences the markdown report flags in red (parameter mismatches excluding Program version, aimec percentage-difference flags, and ppr/pft/pfv plot abs-diff warnings) directly in the run log, without modifying the report itself.


PaulaSp3 (approved):
With this PR: https://github.com/OpenNHM/AvaFrame/pull/1292, no standardtest fails.

PR #1300, opened 2026-06-23
[com1]: call tif check from function 

awirb:


fso42 (approved):
Looks good, only a pytest is missing

fso42 (approved):

fso42:
Standardtests ident (apart from known com5)

PR #1303, opened 2026-06-30
[com1, out3] add debug plot

awirb:


avaframe/out3Plot/outDebugPlots.py
```diff
@@ -374,3 +374,41 @@ def plotVolumeRelease(releaseLine, relThField, releaseLineField):
     fig.colorbar(im1, ax=ax2)
     fig.colorbar(im2, ax=ax3)
     plt.show()
+
+
+def plotParticlesRelease(particles, relRaster, releaseLine, dem, cfg, xyParticlesAll):
```
fso42, line 379: Please add a docstring. maybe even with a short reference to where it make sense to call this from?

fso42 (changes requested):

fso42 (approved):

PR #1305, opened 2026-06-30
[com8]: adjust ini file

awirb:
Due to new .rcf file version and Mot-PSA code update


fso42 (changes requested):
Momentum threshold has a new unit (%) in the new rcf file, so this should be updated too...

fso42 (approved):

PR #1308, opened 2026-07-03
[in3Utils]: cfgUtils handling of `nan`

JuLa96:
When using the OpenNHM-QGIS-Connector on Windows, `in3Utils/cfgUtils.py/convertDF2numerics` could not handle `nan`.

grafik


fso42:
The error message mentions line 951 in cfgUtils (problem with .match) but you changed line 917? Is that on purpose?

JuLa96 (commented):
Works now. No error message anymore.

fso42:
Standardtests ident apart from know com5

PR #1312, opened 2026-07-08
[com6]: add support for `spatialVoellmy` friction model

fso42:
- Add --friction_calibration flag to runCom6RockAvalanche.py (voellmy | spatialVoellmy)          
 - Auto-generate mu/xi rasters from *_spatialVoellmy.shp when spatialVoellmy selected             
 - Fill default values from rock avalanche Voellmy config (mu=0.035, xi=700)                      
 - Error if rasters and shapefile both present (ambiguous)                                        
 - Consolidate two-shapefile input (*_mu.shp / *_xsi.shp) into single *_spatialVoellmy.shp with   
   mu and xsi attributes across spatialVoellmyInputs and runSpatialVoellmyInputs                  
 - Update tests and docs  

## PR Checklist

- [x] I ran `pytest` locally without fails

- [ ] Commits are sensibly squashed and rebased onto latest master
- [ ] Standardtest run without difference (with recompiled cython code)

### DIFF
--- a/avaframe/in3Utils/spatialVoellmyInputs.py
+++ b/avaframe/in3Utils/spatialVoellmyInputs.py
@@ -16,9 +16,9 @@ log = logging.getLogger(__name__)
 
 
 def generateMuXsiRasters(avaDir, cfg):
-    """Generate mu and xi raster files from polygon shapefiles.
+    """Generate mu and xi raster files from a polygon shapefile.
 
-    Reads polygon shapefiles with "mu" and "xsi" attribute fields,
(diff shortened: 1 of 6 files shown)


awirb (changes requested):
if *_mu.asc and *_xi.asc available in Inputs/RASTERS no error is raised (when using runScripts/runSpatialVoellmy) and also files are overwritten if already there with same name

fso42:
Standard tests ok, apart from known hof

PR #1315, opened 2026-07-08
[com1]: Update report com1

awirb:
add info in com1DFA report that projected area and actual area are derived from the release raster and hence are an approximation of the polygon area (deviations are a function of mesh cell size) 

## PR Checklist

- [x] I ran `pytest` locally without fails

- [x] Commits are sensibly squashed and rebased onto latest master
- [x] Standardtest run without difference (with recompiled cython code)

### DIFF
--- a/avaframe/com1DFA/com1DFA.py
+++ b/avaframe/com1DFA/com1DFA.py
@@ -1219,8 +1219,8 @@ def initializeSimulation(cfg, outDir, demOri, inputSimLines, logName):
     relAreaActual = np.sum(relAreaActualList)
     reportAreaInfo = {
         "Release area info": {
-            "Projected Area [m2]": "%.2f" % (relAreaProjected),
-            "Actual Area [m2]": "%.2f" % (relAreaActual),
+            "Projected Area [m2] (raster-based)": "%.0f" % (relAreaProjected),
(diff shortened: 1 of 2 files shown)


avaframe/com1DFA/com1DFA.py
```diff
@@ -1219,8 +1219,8 @@ def initializeSimulation(cfg, outDir, demOri, inputSimLines, logName):
     relAreaActual = np.sum(relAreaActualList)
     reportAreaInfo = {
         "Release area info": {
-            "Projected Area [m2]": "%.2f" % (relAreaProjected),
-            "Actual Area [m2]": "%.2f" % (relAreaActual),
+            "Projected Area [m2] (raster-based)": "%.2f" % (relAreaProjected),
```
fso42: Note from CT: Lets remove those 2 decimal places...

fso42 (changes requested):

fso42 (changes requested):
After discussion with MG and CT: Lets add the 'Model area' i.e. area from volume and release height to the log, but not the report yet. 

-On ice, as the thickness may also be irregular.

JuLa96 (commented):
See comment from @fso42 above

fso42 (approved):

PR #1323, opened 2026-07-31
[ana5]: add general thalweg extension & resampling

### LINKED ISSUE #1232: [com4], [ana5]: Add Flowpy thalweg postprocessing
- [x] Add Scripts for 2D plots
- [x] Add scripts for statistic plots
- [x] try to use mostly Avaframe functions (e.g., for plots)
- [x] use ini file for parameters and the path
- [x] Add docu
- [x] make 2d plot for every pickle in thalwegdata folder?
- [x] mark functions that are not used currently

### LINKED ISSUE #1235: [com4]: thalweg identification
- [x] only compute thalweg and generation lists when the variable is in the output
- [x] make the top extension similar as bottom: depending on the flow direction
- [x] make more points for the extension (with lower distances) to avoid sharp corners
- [x] discuss if it make sense to delete all thalweg points that are not within the raster (see function Path.findLastPointInRaster())
- [x] when resampling the path, we also need to consider resample zDelta (etc..)
- [x] in 2 D plot: represent z and zdelta read from raster
- [x] try to make similar plots as in avaframe (generation thalweg
- [x] add an override configuration of DFAPathGeneration
- [x] only thalweg outputs from com4: centered values: x, y, z, s zDelta, ...
- [x] separate runScript for preparing the path: read values from rasters along thalweg (z, zdelta,...), extension, resampling
- [x] write a runScript for executing FlowPy and preparing the thalweg
- [x] allow many options via config file  for plotting
- [ ] how to deal with weird directions in the end of thalweg due to very few cells?
- [x] add a more detailed doku (in ana5??)

PaulaSp3:
The goal is to derive a thalweg profile from very general thalweg data (containing x and y coordinates).

This changed in each file:

- ana5Utils/DFAPathGeneration.py:
    - black formatter
    - extendProfileTop : add extension option that extends the path to the top similar as the bottom extension that depends on the flow/ path direction
(for option 2, the code is very similar as in extendProfileBottom, should we summarize both in one function with options/different handling for top nd bottom extension?)
    - extendProfileBottom: add possibility to consider lower left center when reading z values from DEM
    - resamplePath : add parameter in configfile that can change the degree of the spline 

- ana5Utils/DFAPathGenerationCfg.ini:
    - add parameter in configfile that can change the degree of the spline (for resampling a path)

- ana5Utils/preparePathGeneral.py
functions to prepare a path/thalweg that has only x and y coordinates: 
    - extend path to top and bottom
    - resample path
    - read z values along path (from DEM)
    - compute s values along path

  
- out3Plot/outCom3Plots.py
    - black formatter
    - generateCom1DFAPathPlot: put the avalanche thalweg location plot into a separate function avalancheThalwegPlot

## PR Checklist

- [ ] I ran `pytest` locally without fails

- [ ] Commits are sensibly squashed and rebased onto latest master
- [ ] Standardtest run without difference (with recompiled cython code)


avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,177 @@
...
+        cfgPathGen["PATH"].getint("extTopOption"),
+        {},
+        profile,
+        dem=demDict,
+        cfg=cfgPathGen["PATH"],
+        considerLLC=True,
+    )
+
+    # extend the bottom
+    profile = DFAPathGeneration.extendProfileBottom(cfgPathGen["PATH"], demDict, profile, considerLLC=True)
```
fso42, line 101: I'm not 100 % sure if the indices that come out are correct or whether they are shifted by one. Can you check? 

- pathExtension sets indStartMassAverage = 1 and indEndMassAverage = size(x) - 2 before the extension calls     
   (preparePathGeneral.py:119-120)                                                                               
 - extendProfileTop (option 2) prepends a single point and extendProfileBottom appends a single one, at least in 
   the normal case                                                                                               
 - The indices are consumed later by resamplePath on the post-extension profile, via s0 and sEnd

And that's then shifted by one? But maybe I did not follow correctly?

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -897,20 +1071,25 @@ def resamplePath(cfg, dem, avaProfile):
...
     indFirst = np.argwhere(avaProfile['s'] >= s0 - resampleDistance/3)[0][0]
     # look for the first point in the extension and take the one before; if the extension is
     # shorter than a resample step, the mass averaged part reaches the last point
-    indEndCandidates = np.argwhere(avaProfile['s'] >= sEnd + resampleDistance/3)
-    indEnd = indEndCandidates[0][0]-1 if len(indEndCandidates) > 0 else np.size(avaProfile['s'])-1
+    try:
+        indEndCandidates = np.argwhere(avaProfile["s"] >= sEnd + resampleDistance / 3)
+        indEnd = indEndCandidates[0][0] - 1 if len(indEndCandidates) > 0 else np.size(avaProfile["s"]) - 1
+        avaProfile['indEndMassAverage'] = indEnd
+    except:
```
fso42: What are you trying to catch here? If there's a real error, it will be masked here...

avaframe/out3Plot/outCom3Plots.py
```diff
@@ -108,102 +130,184 @@ def generateCom1DFAPathPlot(avalancheDir, cfgPath, avaProfileMass, dem, paraboli
...
-    outDir = pathlib.Path(avalancheDir, 'Outputs', 'ana5Utils', 'DFAPath')
+    outFileName = "_".join([simName, "DFAPath"])
+    outDir = pathlib.Path(avalancheDir, "Outputs", "ana5Utils", "DFAPath")
     plt.tight_layout()
-    outPath = pU.saveAndOrPlot({'pathResult': outDir}, outFileName, fig)
+    outPath = pU.saveAndOrPlot({"pathResult": outDir}, outFileName, fig)
     return outPath
+
+
+def avalancheThalwegPlot(ax, fieldRaster, dem, avaProfileMass, splitPoint=None, avalancheDir="", cmapHS=None):
```
fso42, line 264: docstring is missing

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,177 @@
...
+    profileExtended = updateSZProfile(profileExtended, dem)
+    # resample profile/ path and save in an extra dictionary
+    profileExtended = DFAPathGeneration.resamplePath(cfgDFAPath["PATH"], dem, profileExtended)
+    # profileExtended = replaceResampledProfileCore(profileExtended, profileResample)
+
+    # add input parameters to extended profile if they exist
+    for inputPara in ["alpha", "exponent", "zDeltaMax"]:
+        try:
+            profileExtended[inputPara] = profile[inputPara]
+        except:
```
fso42: Should this be a except KeyError? since real error should be shown

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,177 @@
...
+        x, y, s, z of extended and resampled path
+    """
+    # get profile with normalized x and y coordinates and z and s values
+    profileAveraged = updateSZProfile(profile, dem)
+    profileExtended = profileAveraged.copy()
+
+    # skip profile that only contains one point
+    if len(profileAveraged["x"]) <= 2:
+        profileExtended["indStartMassAverage"] = 0
+        profileExtended["indEndMassAverage"] = max(len(profileExtended["x"]) - 1, 1)
```
fso42: Single Points gets '1', but he profile Array only has 0. Do we need to catch this?

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,177 @@
...
+    for inputPara in ["alpha", "exponent", "zDeltaMax"]:
+        try:
+            profileExtended[inputPara] = profile[inputPara]
+        except:
+            continue
+
+    return profileAveraged, profileExtended
+
+
+def replaceResampledProfileCore(profile, profileResample):
```
fso42: Remove, it's dead code an in git anyways

avaframe/ana5Utils/preparePathGeneral.py
PaulaSp3: Where do you mean is the shift by one (in your last line)?

avaframe/ana5Utils/DFAPathGeneration.py
PaulaSp3: I changed this and deleted the `try ... except`.
I got some errors before without that but fixed it also somewhere else ;)

avaframe/ana5Utils/preparePathGeneral.py
PaulaSp3: changed to:
```
    for inputPara in ["alpha", "exponent", "zDeltaMax"]:
        if inputPara in profile.keys():
            profileExtended[inputPara] = profile[inputPara]
```

avaframe/ana5Utils/preparePathGeneral.py
PaulaSp3: True, I've changed to:
`        profileExtended["indEndMassAverage"] = max(len(profileExtended["x"]) - 1, 0)`

avaframe/ana5Utils/preparePathGeneral.py
fso42: So the indeces indStartMassAverage and indEndMassAverage are set before the extension, but then used after the extension. As extendProfileTop prepends one point, every original point shifts by +1. 

The comment says: do not use last point, it should only drop the last point, which now sits at size(x) - 1 and not size(x) - 2 anymore. (But this assumes the top extension happened). 

I'm not sure whether this acually matters....

PaulaSp3: I understand.
The extension functions do not change indStartMassAverage and indEndMassAverage. 
That is why indStartMassAverage = 1 (due to the extra point of top extension).
For the indEndMassAverage, it does not matter (as my results showed) which exact point is used, but for consistency with the comment, I'll change it.

fso42 (approved):

docs/moduleAna5Utils.rst
```diff
@@ -167,6 +176,42 @@ profile.
...
+    (see :ref:`moduleAna5Utils:Resampling`).
+
+The function returns two profiles:
+
+* ``profileAveraged``: the original x, y coordinates with z and s added, left otherwise unmodified.
+* ``profileExtended``: the extended and resampled version of the path, ready to be used as input
+  for modules such as :ref:`moduleCom2AB:com2AB: Alpha Beta Model` or :ref:`moduleAna3AIMEC:ana3AIMEC: Aimec`.
+
+If present in the input, the parameters ``alpha``, ``exponent`` and ``zDeltaMax`` are carried over
+unchanged from the input profile to ``profileExtended``.
```
awirb: as these are scalar values? are these just additional keys in the path dictionary?

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -397,8 +412,8 @@ def extendDFAPath(cfg, avaProfile, dem, particlesIni, fieldPFT=None):
     return avaProfile
 
 
-def extendProfileTop(extTopOption, particlesIni, profile):
-    """ extend the DFA path at the top (release)
+def extendProfileTop(extTopOption, particlesIni, profile, dem=None, cfg=None, considerLLC=False):
+    """extend the DFA path at the top (release)
 
     Either towards the highest point in particlesIni (extTopOption = 0)
     or the point leading to the longest runout (extTopOption = 1)
```
awirb, line 421: add extTopOption=2 here

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -409,10 +424,17 @@ def extendProfileTop(extTopOption, particlesIni, profile):
         decide how to extend towards the top
         if 0, extrapolate towards the highest point in the release
         if 1, extrapolate towards the point leading to the lonest runout
```
awirb:
```suggestion
        if 1, extrapolate towards the point leading to the longest runout
```

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -409,10 +424,17 @@ def extendProfileTop(extTopOption, particlesIni, profile):
...
+        if 2, extrapolate in direction of the thalweg upwards
```
awirb: add info how direction of thalweg is defined? - i.e. nCellsMinExend and nCellsMaxExtend parameters

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -421,51 +443,167 @@ def extendProfileTop(extTopOption, particlesIni, profile):
...
-        zExtTop = particlesIni['z'][indTop]
+        xExtTop = particlesIni["x"][indTop]
+        yExtTop = particlesIni["y"][indTop]
+        zExtTop = particlesIni["z"][indTop]
         ds = ds[indTop]
+    elif extTopOption == 2:
+        if len(profile["x"]) <= 1:
+            # skip computation if thalweg is one point
+            log.warning("Skip top extension of thalweg since profile contains only one point.")
+            return profile
```
awirb, line 485: would it make sense to use this warning outside of the ifs so for all options?

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -421,51 +443,167 @@ def extendProfileTop(extTopOption, particlesIni, profile):
...
+
+        # TODO: for option 2, the code is very similar as in extendProfileBottom, should we summarize both in one function with options/different handling for top nd bottom extension?
```
awirb: I think that would be good

awirb: and if not as there is a lot of the same code in these functions - move the overlapping parts to a separate function and call this one from here and the extendProfileBottom

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -421,51 +443,167 @@ def extendProfileTop(extTopOption, particlesIni, profile):
...
+                factExt = factFirst
+                gamma = factExt * sTotal / vNorm
+                xExtTop = np.array([xFirst - gamma * vDirX])
+                yExtTop = np.array([yFirst - gamma * vDirY])
+                # project on DEM
+                zExtTop, _ = gT.projectOnGrid(
+                    xExtTop, yExtTop, zRaster, csz=csz, xllc=xllcenter, yllc=yllcenter
+                )
+
+            log.info("found extention after %d iterations, precision is %.2f m" % (count, stepSize * sTotal))
```
awirb:
```suggestion
            log.info("found extension after %d iterations, precision is %.2f m" % (count, stepSize * sTotal))
```

avaframe/ana5Utils/DFAPathGeneration.py
```diff
@@ -559,21 +711,27 @@ def extendProfileBottom(cfg, dem, profile):
             xExtBottom = np.array([xLast + gamma * vDirX])
             yExtBottom = np.array([yLast + gamma * vDirY])
             # project on DEM
-            zExtBottom, _ = gT.projectOnGrid(xExtBottom, yExtBottom, zRaster, csz=csz)
-        log.info('found extention after %d iterations, precision is %.2f m' % (count, stepSize * sLast))
+            zExtBottom, _ = gT.projectOnGrid(
+                xExtBottom, yExtBottom, zRaster, csz=csz, xllc=xllcenter, yllc=yllcenter
+            )
+
+        log.info("found extention after %d iterations, precision is %.2f m" % (count, stepSize * sLast))
```
awirb:
```suggestion
        log.info("found extension after %d iterations, precision is %.2f m" % (count, stepSize * sLast))
```

avaframe/ana5Utils/DFAPathGenerationCfg.ini
```diff
@@ -15,6 +15,8 @@ addVelocityInfo = False
 # the path extracted from the DFA simulation is re-sampled
 # re-sampling step size is defined resampleDistance = nCellsResample x cellSize)
 nCellsResample = 10
+# Degree of the spline for splprep.
+kResample = 3
```
awirb, line 19: should this then also be used in the extendDFAPath function and generatePathAndSplitPoint function? to be consistent as there the nCellsResample parameter is read from cfg and default k=3 in the called prepareLine function but so that the parameter introduced here in the cfg file is also effective at the other occurrences?

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,140 @@
+"""
+generate thalweg from x and y coordinates (including extension to top and bottom and resampling)
+"""
+
+import numpy as np
+import logging
+
```
awirb:
```suggestion
# local imports
```

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,140 @@
...
+import avaframe.in3Utils.geoTrans as gT
+from avaframe.ana5Utils import DFAPathGeneration
+
```
awirb:
```suggestion
# create local logger
```

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,140 @@
...
+     3. resample path points
+
+     Parameters
+     -------------
+     profile: dict
+        contains x and y coordinates of thalweg location
+    cfgDFAPath: configparser object
+        configuration for DFA path generation
+    dem: dict
+        elevation model
```
awirb:
```suggestion
        dictionary with header and raster data of elevation model
```

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Returns
+    -------------
+    profileAveraged: dict
+        s and z coordinates are added (x and y original)
```
awirb:
```suggestion
        s and z coordinates are added (x and y original) to input profile
```

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,140 @@
...
+    profileExtended: dict
+        x, y, s, z of extended and resampled path
+    """
+    # get profile with normalized x and y coordinates and z and s values
+    profileAveraged = updateSZProfile(profile, dem)
+    profileExtended = profileAveraged.copy()
```
awirb: is shallow copy intended here? in case profileExtended gets nested objected like a key that is another dict it would also added in profileAveraged - is this the case in line 51 where the resampled path is added as a dict? consider using copy.deepcopy()

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # skip profile that only contains one point
```
awirb: only two or a single point?

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,140 @@
...
+        profileExtended["indStartMassAverage"] = 0
+        profileExtended["indEndMassAverage"] = max(len(profileExtended["x"]) - 1, 0)
+        return profileAveraged, profileExtended
+
+    # if extTopOption == 2, particlesIni are not used!!
+    profileExtended = pathExtension(profileExtended, dem, cfgDFAPath)
+    profileExtended = updateSZProfile(profileExtended, dem)
+    # resample profile/ path and save in an extra dictionary
+    profileExtended = DFAPathGeneration.resamplePath(cfgDFAPath["PATH"], dem, profileExtended)
+    # profileExtended = replaceResampledProfileCore(profileExtended, profileResample)
```
awirb: is this needed?

avaframe/ana5Utils/preparePathGeneral.py
```diff
@@ -0,0 +1,140 @@
...
+        considerLLC=True,
+    )
+
+    # extend the bottom
+    profile = DFAPathGeneration.extendProfileBottom(cfgPathGen["PATH"], demDict, profile, considerLLC=True)
+
+    return profile
+
+
+def updateSZProfile(profile, dem):
```
awirb, line 106: could this be combined or the identical part be moved to a function of what is done in prepareLine from lines 670 onwards to reduce duplicated code?

avaframe/com1DFA/DFAtools.py
```diff
@@ -159,3 +159,43 @@ def scalProd(ux, uy, uz, vx, vy, vz):
...
+    """
+    if vDirZ is None:
+        vDirZ = 0 * np.array([vDirY])
+    else:
+        vDirZ = np.array([vDirZ])
+
+    vDirX, vDirY, vDirZ = normalize(
+        np.array([vDirX]), np.array([vDirY]), vDirZ)
+
+    # summed / averaged directio
```
awirb:
```suggestion
    # summed / averaged direction
```

avaframe/com1DFA/DFAtools.py
```diff
@@ -159,3 +159,43 @@ def scalProd(ux, uy, uz, vx, vy, vz):
...
+    vDirX = np.sum(vDirX)
+    vDirY = np.sum(vDirY)
+    vDirZ = np.sum(vDirZ)
+    vDirX, vDirY, vDirZ = normalize(np.array([vDirX]), np.array([vDirY]), np.array([vDirZ]))
```
awirb, line 200: add comment normalize to get unit vector?

avaframe/out3Plot/outCom1DFA.py
```diff
@@ -239,7 +239,7 @@ def addParticles2Plot(particles, ax, dem, whatS="m", whatC="h", colBarResType=""
     return ax, cb
 
 
-def addDem2Plot(ax, dem, what="slope", extent="", origHeader=False):
+def addDem2Plot(ax, dem, what="slope", extent="", origHeader=False, cmap=None):
```
awirb, line 242: add cmap to docstring

awirb (changes requested):

avaframe/ana5Utils/DFAPathGeneration.py
PaulaSp3: I don't think so, since the thalweg can be extended to top for option 0 and 1, if it's only one point.

avaframe/ana5Utils/DFAPathGenerationCfg.ini
PaulaSp3: I'm not sure, if there was a reason, why this was not used in extendDFAPath().

avaframe/ana5Utils/preparePathGeneral.py
PaulaSp3: Yes, we could, but we would need to define a option, for not considering xllc and yllc. I'm not sure if this is easier code then..

avaframe/ana5Utils/DFAPathGeneration.py
PaulaSp3: The thing is, it is always a bit different..

awirb: and the small differences could not be handled by input parameters in the function that does the part that is overlapping?

avaframe/ana5Utils/DFAPathGenerationCfg.ini
awirb: I think because the default was anyway set to 3 in the function prepareLine, but now that it is a parameter in the ini file it might be confusing that it is there and not applied - at the moment no prob but if for some reason the default in the optional input parameter is adjusted - then it would be different from the one in the ini

avaframe/ana5Utils/DFAPathGeneration.py
awirb: ok I thought if it is just one point maybe the extension is not so meaningful but depending on the topo it could be  - so then no change needed

avaframe/ana5Utils/preparePathGeneral.py
PaulaSp3: Issue https://github.com/OpenNHM/AvaFrame/issues/1335

awirb (approved):

fso42:
Standardtests ok, apart from known com5

